PR #2, opened 2020-05-04
Feature/watch owned resources

mxpaspa:
This is an alteration of the my other pr which includes some refactoring.
Changes Made:
- Takes away the watch on the rhm-operator-secret
- Adds ownership to the razeejob 
- Removes switch/case style since we are no longer accepting requests from different contexts

Remaining Items:

- [x] Change github references to use github.com/redhat-marketplace

- [ ] Test whether adding ownership to the secret after it's found works 

- [ ] Check that I'm returning reconcole.Result{},err in the correct places

- [x] Issue with updating JobState

- [x] Make sure that legacy crd values are set to null/accounted for properly

- [ ] Check that the Spec field and Status field descriptors are present and correct after merge with develop

- [x] Check that updates to: ClusterUUID, RazeeDashUrl, ChildRRS3FileName trigger an update tocorresponding Razee resources


pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -181,24 +156,21 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
 	// Check if the RazeeDeployment instance is being marked for deletion
 	isMarkedForDeletion := instance.GetDeletionTimestamp() != nil
 	if isMarkedForDeletion {
-		if utils.Contains(instance.GetFinalizers(), razeeDeploymentFinalizer) {
-			//Run finalization logic for the razeeDeploymentFinalizer.
+		if utils.Contains(instance.GetFinalizers(), utils.RAZEE_DEPLOYMENT_FINALIZER) {
+			//Run finalization logic for the RAZEE_DEPLOYMENT_FINALIZER.
 			//If it fails, don't remove the finalizer so we can retry during the next reconcile
-			return r.partialUninstall(instance)
+			return r.finalizeRazeeDeployment(instance)
```
zach-source, line 162: Regression, revert to partialUninstall.
```suggestion
			return r.partialUninstall(instance)
```

pkg/utils/helpers.go
```diff
@@ -115,6 +161,100 @@ func AddSecretFieldsToObj(razeeData map[string][]byte) (map[string]string, error
...
+			fmt.Println("AddSecretFieldsToStruct", err)
+			error = err
+		}
+		if err != nil {
+			error = err
+		}
+	}
+
+	missingItems := ContainsMultiple(keys, GetStructFieldsOnBase(marketplacev1alpha1.RazeeConfigurationValues{}))
+	return *razeeStruct, missingItems, error
```
zach-source, line 255: Let's simplify this and just set the struct directly. The struct is hard coded and can not be returned any other way so dynamic code isn't really being used:

```go
obj := &RazeeConfigValues{}
switch key {
case IBM_COS_READER_KEY: obj.IbmCosReaderkey = value // whatever obj you need to set it to
//etc...
}
```

You can use a top level switch to set different fields.

pkg/utils/helpers.go
```diff
@@ -115,6 +161,100 @@ func AddSecretFieldsToObj(razeeData map[string][]byte) (map[string]string, error
...
+	for i := 0; i < aa.NumField(); i++ {
+		field := aa.Type().Field(i)
+		if jsonTag := field.Tag.Get("json"); jsonTag != "" && jsonTag != "-" {
+			if commaIdx := strings.Index(jsonTag, ","); commaIdx > 0 {
+				fieldName := jsonTag[:commaIdx]
+				fieldSlice = append(fieldSlice, string(fieldName))
+			}
+		}
+	}
+	return fieldSlice
```
zach-source, line 216: I like this but the reflect scares me a bit. Can we add some tests?

pkg/utils/helpers.go
```diff
@@ -115,6 +161,100 @@ func AddSecretFieldsToObj(razeeData map[string][]byte) (map[string]string, error
 	return razeeDataObj, error
 }
 
+//TODO: not being used
+func AddSecretFieldsToStruct(razeeData map[string][]byte) (*marketplacev1alpha1.RazeeConfigurationValues, []string, error) {
```
zach-source, line 165: Remove if not in use please.

pkg/utils/helpers.go
```diff
@@ -93,15 +101,53 @@ func RemoveKey(list []string, key string) []string {
...
+	if value, ok := secret.Data[sel.Key]; ok {
+		return value, nil
+	} else if !ok {
+		error = fmt.Errorf("secret %s key %q not in secret", sel.Key, sel.Name)
+	}
+
+	return value, error
+}
+
+func GetDataFromRhmSecret(request reconcile.Request, sel corev1.SecretKeySelector, client client.Client) (error, []byte) {
```
zach-source, line 133: We can make this generic. You've hard coded the secret name. Let's pass in a name and namespace and have it return values.

pkg/utils/helpers.go
```diff
@@ -93,15 +101,53 @@ func RemoveKey(list []string, key string) []string {
 	return newList
 }
 
+func CheckMapKeys(razeeConfigValues map[string]string, referenceList []string) []string {
```
zach-source, line 104: Let's add tests please.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -208,259 +180,270 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
-		if errors.IsNotFound(err) {
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			// report to status that we haven't found the secret
-			reqLogger.Info("Updating RedHatMarketplaceSecretFound")
-			instance.Status.RedHatMarketplaceSecretFound = &redHatMarketplaceSecretFound
-			*instance.Status.RedHatMarketplaceSecretFound = false
-			err = r.client.Status().Update(context.TODO(), instance)
+	// TODO: is there a way to refactor this so reconcileRhmOperatorSecret() doesn't have to be called twice.
```
zach-source, line 190: Calling it twice is fine here. The DeployConfig is created but missing values. You could move it one level above and make it like this:

```go
if instance.Spec.DeployConfig != nil && len(instance.Status.MissingDeploySecretValues) > 0 {
			reqLogger.Info("Missing required razee configuration values")
			return r.reconcileRhmOperatorSecret(*instance, request)
}
```


Place it below the block on line 188.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -208,259 +180,270 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
+			reqLogger.Info("all secret values found")
+
+			//construct the childURL
+			url := fmt.Sprintf("%s/%s/%s/%s", instance.Spec.DeployConfig.IbmCosURL, instance.Spec.DeployConfig.BucketName, instance.Spec.ClusterUUID, instance.Spec.DeployConfig.ChildRSSFIleName)
+			instance.Spec.ChildUrl = &url
+			err = r.client.Update(context.TODO(), instance)
 			if err != nil {
-				reqLogger.Error(err, "Failed to update Status.RedHatMarketplaceSecretFound")
+				reqLogger.Error(err, "Failed to update ChildUrl")
+				return reconcile.Result{}, err
```
zach-source, line 204: We can move this logic into the end of the reconcileRhmOperatorSecret.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -208,259 +180,270 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
-			razeeNamespace.ObjectMeta.Name = targetNamespace
-			return reconcile.Result{RequeueAfter: time.Second * 60}, nil
+			reqLogger.Info("razee namespace does not exist - creating")
+			razeeNamespace.ObjectMeta.Name = *instance.Spec.TargetNamespace
+			err = r.client.Create(context.TODO(), &razeeNamespace)
+			if err != nil {
+				reqLogger.Error(err, "Failed to create razee namespace.")
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{Requeue: true}, nil
```
zach-source, line 226: Regression. Removed namespace creation. Please revert to the code in dev.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -208,259 +180,270 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
+				return reconcile.Result{}, err
+			}
+
+			reqLogger.Info("Updating resource", "resource: ", watchKeeperNonNamespace.Name)
+			err = r.client.Update(context.TODO(), updatedWatchKeeperNonNameSpace)
+			if err != nil {
+				reqLogger.Info(fmt.Sprintf("Failed to update %v", watchKeeperNonNamespace.Name))
+				return reconcile.Result{}, err
+			}
+		}
```
zach-source, line 294: Awesome job here! Really like this.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -841,21 +986,71 @@ func (r *ReconcileRazeeDeployment) finalizeRazeeDeployment(req *marketplacev1alp
...
-	req.SetFinalizers(utils.RemoveKey(req.GetFinalizers(), razeeDeploymentFinalizer))
+	req.SetFinalizers(utils.RemoveKey(req.GetFinalizers(), utils.RAZEE_DEPLOYMENT_FINALIZER))
 	err = r.client.Update(context.TODO(), req)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileRazeeDeployment) MakeRazeeJob(
```
zach-source, line 999: Regression on this - check dev. Code is diff.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -865,98 +1060,111 @@ func (r *ReconcileRazeeDeployment) addFinalizer(razee *marketplacev1alpha1.Razee
...
-			Namespace: namespace,
+			Namespace: *instance.Spec.TargetNamespace,
 		},
-		Data: map[string]string{"RAZEEDASH_URL": rhmOperatorSecretValues.razeeDashUrl, "START_DELAY_MAX": "0"},
+		Data: map[string]string{"RAZEEDASH_URL": instance.Spec.DeployConfig.RazeeDashUrl, "START_DELAY_MAX": "0"},
+	}
+}
+
+// Uses the SecretKeySelector struct to to retrieve byte data from a specified key
+func (r *ReconcileRazeeDeployment) GetDataFromRhmSecret(request reconcile.Request, sel corev1.SecretKeySelector) ([]byte, error) {
```
zach-source, line 1114: Duplicated with utils func? Maybe keep this one and remove it from utils? Hard coded is fine here.

pkg/apis/marketplace/v1alpha1/razeedeployment_types.go
```diff
@@ -54,14 +81,16 @@ type RazeeDeploymentStatus struct {
 	JobState batch.JobStatus `json:"jobState,omitempty"`
 	// MissingValuesFromSecret validates the secret provided has all the correct fields
 	// +operator-sdk:gen-csv:customresourcedefinitions.statusDescriptors=true
-	MissingValuesFromSecret *[]string `json:"missingValuesFromSecret,omitempty"`
```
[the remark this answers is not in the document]
zach-source: Hrm, actually since this is status I'm not sure if removing them will matter. Best to leave them just in case though.

pkg/controller/razeedeployment/razeedeployment_controller.go
mxpaspa: yep I was thinking about exporting from utils at one point.

pkg/controller/razeedeployment/razeedeployment_controller.go
mxpaspa: Hey @zach-source I think this one matches up. I'm just reading the razee job name off the constant in utils, targetNamespace comes from the instance, as does the file source url. .

pkg/apis/marketplace/v1alpha1/razeedeployment_types.go
```diff
@@ -54,14 +81,16 @@ type RazeeDeploymentStatus struct {
...
+	// RazeePrerequestesCreated is the list of configmaps and secrets required to be installed
+	// +operator-sdk:gen-csv:customresourcedefinitions.statusDescriptors=true
+	RazeePrerequisitesCreated []string `json:"razeePrerequisitesCreated,omitempty"`
 	// LocalSecretVarsPopulated informs if the correct local variables are correct set.
+	//TODO: denote as legacy
 	// +operator-sdk:gen-csv:customresourcedefinitions.statusDescriptors=true
 	LocalSecretVarsPopulated *bool `json:"localSecretVarsPopulated,omitempty"`
-	// RazeePrerequestesCreated is the list of configmaps and secrets required to be installed
-	// +operator-sdk:gen-csv:customresourcedefinitions.statusDescriptors=true
-	RazeePrerequisitesCreated *[]string `json:"razeePrerequisitesCreated,omitempty"`
```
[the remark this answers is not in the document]
mxpaspa: I kept this value, but just removed the pointer. Does that need to be marked deprecated ?

pkg/controller/razeedeployment/razeedeployment_controller.go
mxpaspa: Thought about doing that. I was worried about having a partially formed url. As is, I know all the values are there and can create a full url.

pkg/utils/helpers.go
mxpaspa: Not being used.

pkg/utils/helpers.go
mxpaspa: @zach-source can you confirm you want to use this instead of the function in razeedeployment_controller ? this is a dup.

pkg/controller/razeedeployment/razeedeployment_controller.go
zach-source: If you don't have any missing fields and you recalculate it each time the secret is found then by the time you get here the child url will be complete.

pkg/controller/razeedeployment/razeedeployment_controller.go
zach-source: Yep, you're right, my fault on this one.

pkg/apis/marketplace/v1alpha1/razeedeployment_types.go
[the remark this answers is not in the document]
zach-source: Ah, I didn't see you had moved it. You can disregard this one.

pkg/utils/helpers.go
zach-source: Just use the function in razeedeployment and let's clean up dupes/unused code.

pkg/utils/helpers.go
zach-source: Let's remove it then.

zach-source:
We've moved this work to the /owned PR

PR #3, opened 2020-05-04
Feature/  Filter By Namespace

NamSingh:
Created FilterByNamespace function in k8s.go
Allows user to filter and return a ResourceList (of Pods and ServiceMonitors) by Namespace.
* **parameters**: list of namespaces, ResourceList, client
* **returns**: error, ResourceList consisting of Pods and ServiceMonitors filtered by namespace
* added test cases for all 3 scenarios (no namespace, one namespace, multiple namespaces)


pkg/utils/k8s.go
```diff
@@ -207,3 +208,60 @@ func LoadYAML(filename string, i interface{}) (interface{}, error) {
 
 	return genericTypeVal, nil
 }
+
+// filterByNamespace returns a ResourceList of Pods and ServiceMonitors filtered by namespaces
+func FilterByNamespace(namespaces []corev1.Namespace, resources corev1.ResourceList, rClient client.Client) (error, corev1.ResourceList) {
+	var err error
+	if len(namespaces) == 0 {
```
zach-source, line 215: Good check here. Nice defensive programming.

pkg/utils/k8s.go
```diff
@@ -207,3 +208,60 @@ func LoadYAML(filename string, i interface{}) (interface{}, error) {
...
+		}
+	} else {
+		err = errors.New("unexpected length of []namespaces")
+	}
+	return err, resources
+}
+
+// getResources() is a helper function for FilterByNamespace(), it returns a ResourceList in the requested namespaces
+// the namespaces are preset in listOpts
+func getResources(listOpts []client.ListOption, resources corev1.ResourceList, rClient client.Client) (error, corev1.ResourceList) {
```
zach-source, line 245: Let's have getResources take a runtime Object.
```suggestion
func getResources(obj runtime.Object, listOpts []client.ListOption, rClient client.Client) (error, corev1.ResourceList) {
```

Instead of creating a podList or serviceMonitorList you'll just pass in that runtime object (in this case a list), and it will go fetch the results for you. So for example:

```
obj := &PodList
FilterByNamespace(obj, namespaces, rClient)
```
We can drop the resource list. I like the idea but we want to filter on a resource across multiple namespaces and filters.

pkg/utils/k8s_test.go
```diff
@@ -52,3 +92,117 @@ func TestPersistenVolumeClaim(t *testing.T) {
 		t.Errorf("expect %v but got %v", corev1.ReadWriteMany, pvc.Spec.AccessModes[0])
 	}
 }
+
+func TestFilterByNamespace(t *testing.T) {
```
zach-source, line 96: Really like the rigorous testing here 👍

NamSingh:
* FilterByNamespace now takes in a runtime.Object instead
* No longer returns a ResourceList, instead returns the runtime.Object with an updated list of objects
* FilterByNamespace now takes a varadic of ListOption 
* Added a couple more test cases

Note: when passing in multiple namespaces it currently only works with type: PodList, ServiceList, and ServiceMonitorList. I was unable to find a generic way to create/append the runtime.Object from multiple namespaces.; tried passing in []ListOption identifying multiple namespaces but it only returns resources from the last namespace in the array.

PR #26, opened 2020-06-10
fix: delete parent rrs3

jiangchen2019:


pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -1436,15 +1436,17 @@ func (r *ReconcileRazeeDeployment) fullUninstall(
...
+				err := r.client.Delete(context.TODO(), &cr)
 				if err != nil && !errors.IsNotFound(err) {
 					reqLogger.Error(err, "could not delete custom resource", "custom resource", cr)
 				}
 			}
 		}
 	}
 
+	// sleep 5 seconds to let custom resource deletion complete
+	time.Sleep(time.Second * 5)
```
zach-source, line 1448: Actually - this is not what we want to do. We should add this code to line 1442.

```
return reconcile.Result{requeueAfter: time.Second * 5}, nil
```

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -1436,15 +1436,17 @@ func (r *ReconcileRazeeDeployment) fullUninstall(
 		if err == nil {
 			for _, cr := range customResourceList.Items {
 				reqLogger.Info("Deleteing custom resource", "custom resource", cr)
-
-				err := r.client.Delete(context.TODO(), &cr, client.PropagationPolicy(deletePolicy))
+				err := r.client.Delete(context.TODO(), &cr)
 				if err != nil && !errors.IsNotFound(err) {
 					reqLogger.Error(err, "could not delete custom resource", "custom resource", cr)
 				}
```
zach-source, line 1442: At this point we have deleted things, and we need to wait. This will requeue the event in 5 seconds.
```suggestion
				}

                              return reconcile.Result{requeueAfter: time.Second * 5}, nil
```

pkg/controller/razeedeployment/razeedeployment_controller.go
zach-source: Talked it over with Chen, she did it that way over a comment I made about worrying about dependent finalizers not finishing appropriately and re-queue forever since this is the finalizer. Traditionally we requeue this way, but since this is a finalizer we'll just use sleep.

PR #31, opened 2020-06-22
Feature/node controller 

tkishore1:
Issue: https://github.ibm.com/symposium/track-and-plan/issues/6222

What changed:

- Added a controller for node for checking the label [`"razee/watch-resource":"lite"`] and patching it if absent.
- Updated the roles for getting and updating node, clusterversion resources  
- Added a code snippet for patching the label to clusterversion

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -16,6 +16,7 @@ package node
 
 import (
 	"context"
+	"reflect"
 
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -24,7 +25,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -62,21 +62,16 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	labelPreds := []predicate.Predicate{
 		predicate.Funcs{
 			UpdateFunc: func(evt event.UpdateEvent) bool {
-				// get node with the update event name
-				rNode := r.(*ReconcileNode)
-				instance := &corev1.Node{}
-				rNode.client.Get(context.Background(), types.NamespacedName{Name: evt.MetaOld.GetName()}, instance)
-				watchResourceTagLabel, watchOk := instance.GetLabels()[watchResourceTag]
-
-				return !(watchOk && watchResourceTagLabel == watchResourceValue)
+				watchResourceTag, ok := evt.MetaNew.GetLabels()[watchResourceTag]
+				return !(ok && watchResourceTag == watchResourceValue)
+			},
+			CreateFunc: func(evt event.CreateEvent) bool {
+				watchResourceTag, ok := evt.Meta.GetLabels()[watchResourceTag]
+				return !(ok && watchResourceTag == watchResourceValue)
 			},
 			GenericFunc: func(evt event.GenericEvent) bool {
-				rNode := r.(*ReconcileNode)
-				instance := &corev1.Node{}
-				rNode.client.Get(context.Background(), types.NamespacedName{Name: evt.Meta.GetName()}, instance)
-				watchResourceTagLabel, watchOk := instance.GetLabels()[watchResourceTag]
-
-				return !(watchOk && watchResourceTagLabel == watchResourceValue)
+				watchResourceTag, ok := evt.Meta.GetLabels()[watchResourceTag]
+				return !(ok && watchResourceTag == watchResourceValue)
 			},
 		},
 	}
@@ -119,7 +114,6 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result, 
 			reqLogger.Error(err, "node does not exist")
 			return reconcile.Result{}, nil
 		}
-		// Error reading the object - requeue the request.
 		reqLogger.Error(err, "Failed to get node")
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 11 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,140 @@
...
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("node-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	labelPreds := []predicate.Predicate{
+		predicate.Funcs{
+			UpdateFunc: func(evt event.UpdateEvent) bool {
```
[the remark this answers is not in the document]
zach-source: I figure that Chen talked with you about this. Just for anyone who may come across it. The watch functions don't need to query for the type and run client commands. Ideally the watches are lean and fast because these get called for every inbound event. You would only use the labels on the evt itself or the information the evt provides to provide watch filters.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -1064,6 +1064,38 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
+			Name: "version",
+		}, clusterVersion)
+		if err != nil {
+			reqLogger.Error(err, "Failed to retrieve clusterversion resource")
+			return reconcile.Result{}, err
+		}
+
+		reqLogger.V(0).Info("Found clusterversion resource")
+		clusterVersionLabels := clusterVersion.GetLabels()
+		if !reflect.DeepEqual(clusterVersionLabels, RAZEE_WATCH_KEEPER_LABELS) || clusterVersionLabels == nil {
```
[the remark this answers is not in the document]
zach-source: Yea, you would use DeepEqual to compare old with new. You can do this to see if you need to update.

```suggestion
                 originalList := clusterVersion.GetLabels().DeepCopy()
		if !reflect.DeepEqual(clusterVersionLabels, originalList) {
```

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+func testMultipleNodes(t *testing.T) {
+	t.Parallel()
+	reconcilerTest := NewReconcilerTest(setup, nodeLabelsDiff.DeepCopyObject(), node.DeepCopyObject(), nodeLabelsAbsent.DeepCopyObject())
+	reconcilerTest.TestAll(t,
+		// Reconcile to create obj
+		ReconcileStep(opts,
+			ReconcileWithExpectedResults(DoneResult)),
+		// List and check results
+		ListStep([]StepOption{},
```
zach-source, line 209: Nice use of the testing tools!

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	labels := instance.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[watchResourceTag] = watchResourceValue
+	instance.SetLabels(labels)
+
+	if err := r.client.Update(context.TODO(), instance); err != nil {
```
zach-source, line 134: Please add an `if` condition to check if an update is required. Even if we filter via the watch, we should still have an `if` check here.

pkg/controller/node/node_controller.go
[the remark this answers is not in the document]
tkishore1: Got it thanks.

pkg/controller/razeedeployment/razeedeployment_controller.go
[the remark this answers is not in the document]
tkishore1: @zach-source I should check even for empty values right? This `if !reflect.DeepEqual(clusterVersionLabels, originalList)` return always true if there are no labels present. 

Do you want me to update other resources ( infrastructure and console) as well ?

pkg/controller/node/node_controller.go
tkishore1: I can add the condition but according to the watch only the nodes which doesn't have the label `razee/watch-resource ` gets updated right. So is there any chance where watch fails and it gets here even the label present ?

zach-source: There isn't a chance, but adding an if protects future devs who may change the watch conditions.

tkishore1: Sounds good 👍🏻

pkg/controller/razeedeployment/razeedeployment_controller.go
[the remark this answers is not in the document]
zach-source: The Reconcile loop get's triggered a lot. You're always looking at current state and moving to desired state. So typically the way I do this is this kind of pattern.

1. Create a copy of the original.
2. Modify the original with any potential additions you would make.
3. Compare the two.
4a. If not equal, update and requeue.
4b. Else continue.

So code like this: 
```go
originalLabels := clusterVersion.GetLabels().DeepCopy()
clusterVersionLabels := clusterVersion.GetLabels()
clusterVersionLabels[razeeWatchTag] = razeeWatchTagValue
if !reflect.DeepEqual(clusterVersionLabels, originalList) {
   client.Update(ctx, clusterVersion)
}
```

Will only update if you add the tag.

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -1021,16 +1023,20 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
 		reqLogger.V(0).Info("Found Console resource")
+		consoleOriginalLabels := console.DeepCopy().GetLabels()
 		consoleLabels := console.GetLabels()
-
-		if !reflect.DeepEqual(consoleLabels, RAZEE_WATCH_KEEPER_LABELS) || consoleLabels == nil {
+		if consoleLabels == nil {
+			consoleLabels = make(map[string]string)
+		}
+		consoleLabels[razeeWatchTag] = razeeWatchTagValue
+		if !reflect.DeepEqual(consoleLabels, consoleOriginalLabels) {
```
jiangchen2019, line 1032: Due to the change, line 1033 need be `console.SetLabels(consoleLabels)`

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -1051,19 +1057,61 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
 
 		reqLogger.V(0).Info("Found Infrastructure resource")
+		infrastructureOriginalLabels := infrastructureResource.DeepCopy().GetLabels()
 		infrastructureLabels := infrastructureResource.GetLabels()
-		if !reflect.DeepEqual(infrastructureLabels, RAZEE_WATCH_KEEPER_LABELS) || infrastructureLabels == nil {
+		if infrastructureLabels == nil {
+			infrastructureLabels = make(map[string]string)
+		}
+		infrastructureLabels[razeeWatchTag] = razeeWatchTagValue
+		if !reflect.DeepEqual(infrastructureLabels, infrastructureOriginalLabels) {
```
jiangchen2019, line 1066: line 1067, infrastructureResource.SetLabels(infrastructureLabels)

pkg/controller/razeedeployment/razeedeployment_controller.go
```diff
@@ -1051,19 +1057,61 @@ func (r *ReconcileRazeeDeployment) Reconcile(request reconcile.Request) (reconci
...
+		reqLogger.V(0).Info("Found clusterversion resource")
+		clusterVersionOriginalLabels := clusterVersion.DeepCopy().GetLabels()
+		clusterVersionLabels := clusterVersion.GetLabels()
+		if clusterVersionLabels == nil {
+			clusterVersionLabels = make(map[string]string)
+		}
+		clusterVersionLabels[razeeWatchTag] = razeeWatchTagValue
+		if !reflect.DeepEqual(clusterVersionLabels, clusterVersionOriginalLabels) {
+
+			clusterVersion.SetLabels(RAZEE_WATCH_KEEPER_LABELS)
```
jiangchen2019, line 1103: `clusterVersion.SetLabels(clusterVersionLabels)`

PR #35, opened 2020-06-25
Revert "Feature/uninstall subscription (#28)"

zach-source:
Bug found, reverting commit. This reverts commit 77046a8d455875c7211bad65a6e141cefece65e6.

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -94,13 +94,10 @@ roles:
           - operators.coreos.com
         resources:
           - subscriptions
-          - installplans
-          - clusterserviceversions
         verbs:
           - get
           - list
           - watch
-          - delete
       - apiGroups:
           - operators.coreos.com
         resources:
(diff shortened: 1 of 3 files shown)


deploy/chart/values.yaml
```diff
@@ -156,6 +153,21 @@ roles:
           - list
           - patch
           - update
+      - apiGroups:
+          - apiextensions.k8s.io
+        resources:
+          - customresourcedefinitions
```
tkishore1, line 159: Do we need this RBAC permissions on customresourcedefinitions?

jiangchen2019: We don't need it. But since this is a revert, I think we can just keep it now and remove it in a later PR

zach-source: Exactly, not needed but it's a revert of the PR. We'll just re-open the PR and add the changes back with the fix.

PR #39, opened 2020-06-29
Feature/catalog source flag

NamSingh:
[Issue at hand](https://github.ibm.com/symposium/track-and-plan/issues/7731)

**What Changed**
* User can decide whether to install IBM Catalog Source via flags or `MarketplaceConfig.Spec.InstallIBMCatalogSource` (priority given to MarketplaceConfig.Spec)
* If there is no `MarketplaceConfig.Spec.InstallIBMCatalogSource` is provided, will use the flag instead.
* Flag `IBMCatalogSource` is default to true
* If true: install IBM Catalog Source
* If false: don't install IBM Catalog Source, and delete existing one (if it exists)


sa.kubeconfig
```diff
@@ -0,0 +1,20 @@
+
```
tkishore1, line 1: I think we should be adding this file to gitignore as it was auto generated.

zach-source: yea agreed. We can remove it later.

PR #40, opened 2020-06-30
Feature/add CSV controller

tkishore1:
Issue: https://github.ibm.com/symposium/track-and-plan/issues/7517

What changed:

Added a controller for CSV for adding label ["razee/watch-resource":"lite"] if subscription has rhm operator label
Updated the roles for ClusterServiceVersion resources
Not attaching razee watch label to the copied CSVs for allNamespace subscriptions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ setup-operator-sdk-run: ## Create ns, crds, sa, role, and rolebinding before ope
 	- . ./scripts/operator_sdk_sa_kubeconfig.sh $(CLUSTER_SERVER) $(NAMESPACE) $(SERVICE_ACCOUNT)
 
 operator-sdk-run: ## Run operator locally outside the cluster during development cycle
-	operator-sdk run local --watch-namespace="" --kubeconfig=./sa.kubeconfig
+	operator-sdk run --local --watch-namespace="" --kubeconfig=./sa.kubeconfig
 
 ##@ Manual Testing
 
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ setup-operator-sdk-run: ## Create ns, crds, sa, role, and rolebinding before ope
 	- . ./scripts/operator_sdk_sa_kubeconfig.sh $(CLUSTER_SERVER) $(NAMESPACE) $(SERVICE_ACCOUNT)
 
 operator-sdk-run: ## Run operator locally outside the cluster during development cycle
-	operator-sdk run --local --watch-namespace="" --kubeconfig=./sa.kubeconfig
+	operator-sdk run local --watch-namespace="" --kubeconfig=./sa.kubeconfig
 
 ##@ Manual Testing
 
--- a/pkg/controller/clusterserviceversion/clusterserviceversion_controller.go
+++ b/pkg/controller/clusterserviceversion/clusterserviceversion_controller.go
@@ -1,0 +1,160 @@
+// Copyright 2020 IBM Corp.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clusterserviceversion
+
+import (
+	"context"
+	"reflect"
+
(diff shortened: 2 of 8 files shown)


pkg/controller/clusterserviceversion/clusterserviceversion_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	c, err := controller.New("clusterserviceversion-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	labelPreds := []predicate.Predicate{
+		predicate.Funcs{
+			UpdateFunc: func(evt event.UpdateEvent) bool {
+				return false
+			},
```
zach-source, line 67: Also need to include a DeleteFunc predicate here too because it'll enqueue it all by default.
```suggestion
			UpdateFunc: func(evt event.UpdateEvent) bool {
				return false
			},
			DeleteFunc: func(evt event.UpdateEvent) bool {
				return false
			},
```

pkg/controller/clusterserviceversion/clusterserviceversion_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a ClusterServiceVersion object and makes changes based on the state read
+// and what is in the ClusterServiceVersion.Spec
+func (r *ReconcileClusterServiceVersion) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Name", request.Name)
+	reqLogger.Info("Reconciling ClusterServiceVersion")
+	// Fetch the ClusterServiceVersion instance
+	CSV := &olmv1alpha1.ClusterServiceVersion{}
```
zach-source, line 105:
```suggestion
	csv := &olmv1alpha1.ClusterServiceVersion{}
```
Style, should be lower case.

pkg/controller/clusterserviceversion/clusterserviceversion_controller.go
```diff
@@ -0,0 +1,160 @@
...
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("clusterserviceversion-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	labelPreds := []predicate.Predicate{
+		predicate.Funcs{
+			UpdateFunc: func(evt event.UpdateEvent) bool {
```
zach-source, line 65: Do you want to queue on Update as well? We should also filter on if the label is already present or not.

tkishore1: AC in story had to apply labels only when a new CSV been created, do I need filters even on UpdateFunc?

zach-source: Just worry that we may miss one if it errors or if we're updating/down when a create happens.

tkishore1: Yeah that makes sense so will add filters on the updateEvent

Makefile
```diff
@@ -153,7 +153,7 @@ setup-operator-sdk-run: ## Create ns, crds, sa, role, and rolebinding before ope
 	- . ./scripts/operator_sdk_sa_kubeconfig.sh $(CLUSTER_SERVER) $(NAMESPACE) $(SERVICE_ACCOUNT)
 
 operator-sdk-run: ## Run operator locally outside the cluster during development cycle
-	operator-sdk run local --watch-namespace="" --kubeconfig=./sa.kubeconfig
+	operator-sdk run --local --watch-namespace="" --kubeconfig=./sa.kubeconfig
```
NamSingh, line 156: I suggest keeping the original line as `--local` is deprecated in newer versions of operator-sdk.
```suggestion
	operator-sdk run local --watch-namespace="" --kubeconfig=./sa.kubeconfig
```

zach-source: Ah didn't realize that. Kishore just make sure you have the latest operator-sdk.

zach-source:
> Looks good for the most part, only suggestion I have is to keep the original makefile command, as `operator-sdk run --local` is deprecated in newer versions.

@tkishore1 

I've rolled back that one command - just need to update your local.

tkishore1:
> > Looks good for the most part, only suggestion I have is to keep the original makefile command, as `operator-sdk run --local` is deprecated in newer versions.
> 
> @tkishore1
> 
> I've rolled back that one command - just need to update your local.



> > Looks good for the most part, only suggestion I have is to keep the original makefile command, as `operator-sdk run --local` is deprecated in newer versions.
> 
> @tkishore1
> 
> I've rolled back that one command - just need to update your local.

Thank you. Updated my operator-sdk version

PR #63, opened 2020-08-26
Bugfix/meterbase name

NamSingh:
**What Changed**

* Will no longer create multiple prometheus-meterbase instances with new MarketplaceConfigs. 
* Will use the asset names for prometheus instead of MarketplaceConfig's name.

_Note:_
* Changed the name from `proemtheus-meterbase-0` to `prometheus-rhm-meterbase`
* Changed MarketplaceConfig name from `marketplaceconfig` to `rhm-marketplaceconfig`
* Tested, still able to create and track meterdefinitions across namespaces (port forwarding: `rhm-state-metrics` and `prometheus-rhm-meterbase)`


zach-source (approved):

zach-source (changes requested):
Unfortunately I didn't catch it on initial pass. This does a few things I want to avoid. We'll need to rework it a bit.

1) We can't change marketplaceconfig name, that's created by an install script.
2) Changing the name of prometheus causes the PVCs to be recreated, this would cause a lost of data in a production environment.

When making changes to the operator - we have to think about what is currently happening and what would happen. If it's drastically different, we have to write migrations. In this case, I'd prefer to just close this PR for now and readdress later.

zach-source:
Closing for now, this is really a nit and something we can readdress in the future.

PR #69, opened 2020-09-16
fix: meterreport pod fails if there are no meterdefs found

mxpaspa:
fix for issue: [https://github.ibm.com/symposium/track-and-plan/issues/10136)](url)


pkg/reporter/reporter.go
```diff
@@ -99,7 +99,7 @@ func (r *MarketplaceReporter) CollectMetrics(ctxIn context.Context) (map[MetricK
 	var resultsMapMutex sync.Mutex
 
 	if len(r.meterDefinitions) == 0 {
-		return resultsMap, []error{}, errors.Wrap(ErrNoMeterDefinitionsFound, "no meterDefs found")
+		logger.Info("no meterdefs found")
```
zach-source, line 102: Let's still return here so we don't go down and do anything else. 
```suggestion
		logger.Info("no meterdefs found")
		return resultsMap, []error{}, nil
```

scripts/operator_sdk_sa_kubeconfig.sh
```diff
@@ -1,34 +0,0 @@
-#!/bin/sh
```
zach-source, line 1: Let's add this file back please.

zach-source (changes requested):

zach-source (dismissed):

zach-source (approved):

PR #70, opened 2020-09-16
Check for expired tokens and restart pods if they're bad

zach-source:
To test:

1. Install operator.
2. Install a marketplace config.
3. Wait until all the pods are available.
4. Uninstall operator.
5. See pods start to fail in openshift-redhat-marketplace.
6. Install operator.
7. See pods get removed and recreated and work again.


mukulpanwar (approved):
Tested, Looking good.

mukulpanwar (approved):
Looking good

PR #72, opened 2020-10-02
Feature/watch keeper 6.6

mxpaspa:
Updates the watch-keeper deployment to use 6.6


deploy/crds/marketplace.redhat.com_meterbases_crd.yaml
```diff
@@ -59,8 +59,9 @@ spec:
               description: Prometheus deployment configuration.
               properties:
                 replicas:
-                  description: 'OPTIONAL: Replicas for the deployment. Used primarily
-                    when running metering to CRC'
+                  description: Replicas defines the number of desired replicas for
```
mxpaspa, line 62: Diff from merging with develop

deploy/crds/marketplace.redhat.com_meterbases_crd.yaml
```diff
@@ -103,8 +104,9 @@ spec:
                         Default is "" i.e. default.
                       type: string
                     emptyDir:
-                      description: EmptyDir is a temporary storage type that is used
-                        to
+                      description: EmptyDir is a temporary storage type that gets
```
mxpaspa, line 107: Diff from merging with develop

deploy/crds/marketplace.redhat.com_v1alpha1_marketplaceconfig_cr.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: marketplace.redhat.com/v1alpha1
 kind: MarketplaceConfig
 metadata:
-  name: rhm-marketplaceconfig
+  name: marketplaceconfig
```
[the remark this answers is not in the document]
zach-source: This shouldn't be changed - it's ok that it's the wrong name. The problem is that the operator scanning fails if it's the name we use for standard deployment. If it's this name, the scan breaks.
```suggestion
  # NOTE: Named this way to prevent an error in scanning
  name: rhm-marketplaceconfig
```

zach-source (changes requested):
Looks good, just revert that one file and add the note.

jiangchen2019 (approved):

zach-source (approved):

PR #73, opened 2020-10-02
ensure default storage class is set in prometheus cr

sgrube:
ensure that a default service class is set in the prometheus CR generated in the meterbase controller


zach-source:
Closing this one @sgrube because bugfix is case sensitive (so builds can run). Opened a new one with a newly named branch.

PR #81, opened 2020-10-15
Feature/additional fields on meterdef

mxpaspa:


pkg/reporter/reporter_test.go
```diff
@@ -255,11 +226,82 @@ var _ = PDescribe("Reporter", func() {
 
 		close(done)
 	}, 20)
+
+	Describe("AdditionalFields logic", func() {
```
zach-source, line 230: Don't need a describe here, should be top level.

pkg/reporter/reporter_test.go
```diff
@@ -255,11 +226,82 @@ var _ = PDescribe("Reporter", func() {
...
+				additionalFields := Keys{
+					"test_field_1": Equal("test-value-1"),
+					"test_field_2": Equal("test-value-2"),
+				}
+
+				for label, matcher := range additionalFields {
+					additionalLabels[label] = matcher
+				}
+
+				sut.meterDefinitions[0].Spec.Workloads[0].MetricLabels[0].Query = "rate(rpc_durations_seconds_count{test_field_1=test-value-1,test_field_2=test-value-2}[5m])*100"
```
zach-source, line 243: Preconditions should be set in a BeforeEach() - You can do that in the when before the It.

pkg/reporter/reporter_test.go
```diff
@@ -255,11 +226,82 @@ var _ = PDescribe("Reporter", func() {
...
+				runTestOnFiles(files, expectedFields)
+				close(done)
+			}, 20)
+		})
+
+		When("AdditionalFields aren't found in the meterdef query", func() {
+			It("Should report an error that indicates the mismatchted fields", func(done Done) {
+
+				sut.meterDefinitions[0].Spec.Workloads[0].MetricLabels[0].Query = "rate(rpc_durations_seconds_count{test_field_1=test-value-1,test_field_2=test-value-2}[5m])*100"
+				sut.meterDefinitions[0].Spec.Workloads[0].MetricLabels[0].AdditionalFields = []string{"wrong_field_1", "wrong_field_2", "test_field_2"}
```
zach-source, line 272: Preconditions should be in a BeforeEach

pkg/reporter/reporter_test.go
```diff
@@ -199,46 +207,9 @@ var _ = PDescribe("Reporter", func() {
...
-							}),
-							"version": Equal("v1"),
-						}),
-					}),
-				}))
-			}
-		}
+		Expect(len(files)).To(Equal(2), "length of files")
+
+		runTestOnFiles(files, expectedFields)
```
zach-source, line 212: This test is failing because of a race condition. If the additional fields tests run before, it sets the additionalFields memory.

pkg/reporter/reporter_test.go
```diff
@@ -51,12 +51,46 @@ var _ = PDescribe("Reporter", func() {
...
+		"version": Equal(""),
+	}
+
+	expectedFields := Keys{
+		"report_slice_id": Not(BeEmpty()),
+		"metrics": MatchElements(id, AllowDuplicates, Elements{
+			"row": MatchAllKeys(expectedMetricRow),
+		}),
+	}
+
```
zach-source, line 93: This is a good idea - but each variables needs to be a var and they need to be set in a BeforeEach, if you don't bad things can happen like race conditions.

Test are randomly failing because of this.

pkg/reporter/reporter_test.go
zach-source, line 243: The prometheus query isn't correct. Inner fields are like `test_field_1="foo"`

pkg/reporter/reporter.go
```diff
@@ -481,3 +516,40 @@ func wgWait(ctx context.Context, processName string, maxRoutines int, done chan
...
+func addAdditionalFieldKVPairsToMap(labelPairs []string, additionalField string ,additionalFieldMap map[string]string){
+	for _, keyValuePairString := range labelPairs {
+		keyValuePairArray := strings.Split(keyValuePairString, "=")
+		if additionalField == keyValuePairArray[0] {
+			additionalFieldMap[keyValuePairArray[0]] = keyValuePairArray[1]
+		}
+	}
+}
+
+func processAdditionalFields(queryLabel string, originalQueryString string, additionalFieldsFromMeterDef []string, additionalFieldMap map[string]string) []error {
```
zach-source, line 540: This function is setting the value off of the query. We want the additional field to be carried over through the query, not necessarily set off of the query. You would add the field as an additionalField to the downstream so it can be pulled out of the result.

zach-source (changes requested):
Unfortunately, noticed some issues. Biggest appears to be using the query to set the value to be included. What we want to do is modify the [query](https://github.com/redhat-marketplace/redhat-marketplace-operator/blob/develop/pkg/reporter/query.go#L42) functions to include the additional field. Validation of it being on the query is fine, but your prometheus string isn't. I'm worried that may introduce a bug, so let's not worry about validation.

Let's have the additional field values from the CR get passed downstream, then we can pass that into the query to be added to final produced query.

For instance, for a pod. Say I wanted field foo

https://github.com/redhat-marketplace/redhat-marketplace-operator/blob/develop/pkg/reporter/query.go#L58
would be `* on(pod,namespace,foo) group_right`

and

https://github.com/redhat-marketplace/redhat-marketplace-operator/blob/develop/pkg/reporter/query.go#L73
would be `%v by (pod,namespace,foo)`

would add it as an extra param.

pkg/reporter/reporter.go
```diff
@@ -481,3 +516,40 @@ func wgWait(ctx context.Context, processName string, maxRoutines int, done chan
...
+		if !strings.Contains(parsedQueryString, additionalField) {
+			msg := fmt.Sprintf("Query doesn't contain a label key for additionalField: %s", additionalField)
+			queryValidationErrors = append(queryValidationErrors, errors.New(msg))
+		} else {
+			addAdditionalFieldKVPairsToMap(labelPairs,additionalField,additionalFieldMap)
+		}
+	}
+
+	return queryValidationErrors
+}
```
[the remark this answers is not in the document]
mxpaspa: sure thing

pkg/reporter/reporter_test.go
mxpaspa: Ok

pkg/reporter/reporter_test.go
mxpaspa: yep, will re-work

zach-source:
This was reworked in a separate branch.

PR #90, opened 2020-11-09
fix(reporter): more unique meter ID generation

zach-source:
This PR will add a few fields to the meter report ID generation to make it more unique. Currently reruns will not happen if the meter report was updated.


mxpaspa (approved):
@zach-source ran a report locally and confirmed that the reportID is getting generated and the metadata contains the new fields you added.

PR #120, opened 2021-01-27
Feature/trim rbac

dacleyra:
Remove the operator(manager) role & rolebinding from classic
Implement the operator(manager) roles via the kubebuilder annotations for each controller & webhook: generates v2/config/rbac.role.yaml

Scope RBAC and watch predicate to cluster or namespace for marketplace.redhat.com
get;list;watch must be cluster scope for manager cache
create;update;patch;delete scoped to namespace where possible

marketplaceconfigs - namespace
meterbases - namespace, but did not implement predicate filter, unsure of correct implementation
meterdefinitions - cluster
meterreports - namespace
razeedeployments - namespace
remoteresources3s - namespace


v2/controllers/marketplace/marketplaceconfig_controller.go
```diff
@@ -71,6 +74,21 @@ type MarketplaceConfigReconciler struct {
...
+// +kubebuilder:rbac:groups=marketplace.redhat.com,resources=razeedeployments,verbs=get;list;watch
+// +kubebuilder:rbac:groups=marketplace.redhat.com,namespace=system,resources=razeedeployments,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=marketplace.redhat.com,resources=meterbases,verbs=get;list;watch
+// +kubebuilder:rbac:groups=marketplace.redhat.com,namespace=system,resources=meterbases,verbs=get;list;watch;create;update;patch;delete
+
+// +kubebuilder:rbac:groups="operators.coreos.com",resources=operatorsources;catalogsources,verbs=get;list;watch
+
+// Could be namespace=openshift-marketplace but problematic for kustomize
+// +kubebuilder:rbac:groups="operators.coreos.com",resources=operatorsources,resourceNames=redhat-marketplace,verbs=create
+// +kubebuilder:rbac:groups="operators.coreos.com",resources=catalogsources,resourceNames=ibm-operator-catalog;opencloud-operators,verbs=create;delete
```
pgodowski, line 90: under which condition RHM operator would **delete** IBM operator catalog and/or opencloud catalog?

v2/controllers/marketplace/clusterregistration_controller.go
```diff
@@ -53,6 +53,13 @@ type ClusterRegistrationReconciler struct {
 	Log    logr.Logger
 }
 
+// +kubebuilder:rbac:groups="",resources=secret,verbs=get;list;watch
```
pgodowski, line 56: is read access to any Secret really required? perhaps can be restricted to specific namespace?

v2/controllers/marketplace/marketplaceconfig_controller.go
```diff
@@ -71,6 +74,21 @@ type MarketplaceConfigReconciler struct {
 	cfg    config.OperatorConfig
 }
 
+// +kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=secret,verbs=get;list;watch
```
pgodowski, line 78: can the cluster-wide access to Secrets be restricted to specific namespace or specific Secret

v2/controllers/marketplace/razeedeployment_controller.go
```diff
@@ -200,6 +217,26 @@ func (r *RazeeDeploymentReconciler) SetupWithManager(mgr manager.Manager) error
 		Complete(r)
 }
 
+// +kubebuilder:rbac:groups="",resources=configmaps;pods;secrets,verbs=get;list;watch
```
pgodowski, line 220: is there a way to request those permissions only if Razee is enabled?
`instance.Spec.Enabled`

v2/controllers/marketplace/marketplaceconfig_controller.go
dacleyra: The operator manager is set to cluster scope, since there are several resources that must be watched at that scope.
The manager's cache function will throw errors if it does not have cluster get;list;watch on any resources types we watch, or otherwise do structured interaction with.

v2/controllers/marketplace/razeedeployment_controller.go
zach-source: Unfortunately no. Permissions ship with the operator. To change permissions requires permissions to modify the roles.

pgodowski: is it fair to say that the global access to secrets, namespaces and nodes is mostly due to Razee?

zach-source: Yes. That level of permission is mostly related to Razee. There are some cases where wider permissions are required but can be scoped down to resources or where I can resource names. 

I think I can scope down the gets as well. Global secret get is probably not required. When we merge these stories I'll take a hard look at everything again.

pgodowski: Right. Global get for all secrets is what needs to be narrowed. Thx!

zach-source (approved):

PR #121, opened 2021-01-28
Feature/resource requests

dacleyra:
Set resources requests & limits

Update the prometheus CR reloader containers with defined resources, such that they can be overriden by factory
Add resources requests to operator proxy containers
Add operator env's to override resource limits of operands
Add factory funcs for setting resources requests & limits on operand containers


zach-source (commented):
Old - clearing review request

PR #129, opened 2021-02-08
remove 'operator' from display name

mxpaspa:
5th issue from: [https://github.ibm.com/symposium/track-and-plan/issues/14332](url)

```
[ERROR] deploy/crds/marketplace.redhat.com_marketplaceconfigs_crd.yaml: ["app.kubernetes.io/instance" "app.kubernetes.io/managed-by" "app.kubernetes.io/name"] not defined under metadata.labels (RequiredMetadataLabelsDefined) --> see references https://github.ibm.com/IBMPrivateCloud/content-verification/blob/master/docs/rules/RequiredMetadataLabelsDefined.md
cert requires these attributes on everything … initially its only seeing the static files, but once the pipeline is rolling it will catch the deployments as well
use redhat-marketplace, add to lower level objects in factory and all objects in v2/config
```


dacleyra (approved):
lgtm

PR #133, opened 2021-02-11
basicAuth for kubeVersion lt 1.20.0

dacleyra:
Use basicAuth and don't use TokenRequestProject until it is GA in kubeVersion 1.20.0+
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

### DIFF
--- a/v2/controllers/marketplace/meterbase_controller.go
+++ b/v2/controllers/marketplace/meterbase_controller.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/gotidy/ptr"
+	"github.com/redhat-marketplace/redhat-marketplace-operator/v2/pkg/config"
 	"github.com/redhat-marketplace/redhat-marketplace-operator/v2/pkg/inject"
 	"github.com/redhat-marketplace/redhat-marketplace-operator/v2/pkg/manifests"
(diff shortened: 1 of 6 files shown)


v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -583,6 +584,17 @@ func (r *MeterBaseReconciler) generateExpectedDates(endTime time.Time, loc *time
 }
 
 func (r *MeterBaseReconciler) newMeterReport(namespace string, startTime time.Time, endTime time.Time, meterReportName string, instance *marketplacev1alpha1.MeterBase, prometheusServiceName string) *marketplacev1alpha1.MeterReport {
+	// If kubeVersion < 1.20 TokenRequest and TokenRequestProjection are beta and not assumed available, use basicAuth (https port). Logical default if unknown kubeVersion parse failure
```
zach-source, line 587: We'll need to rework and merge with all the rework for prometheus. Using 1.20 as a cut off point makes sense from the Feature Gate perspective but many servers have this enabled by default before 1.20.

zach-source (changes requested):

zach-source:
Let's look at the changes for prometheus and incorporate this with it. We'll need to find a better way to check if the feature gate is enabled or available to use outside of just version since it is an optional feature. It is a better method of authenticating. Additionally we need to expand this for our meterdef support.

PR #146, opened 2021-03-03
"Vanilla" Kubernetes - Certificate Issuer Controller

MichalWrobel3:
It creates CA and certificates signed with it for configmaps with annotation `service.beta.openshift.io/inject-cabundle` and generated secrets:
`prometheus-operator-tls`
`rhm-metric-state-tls`
`rhm-prometheus-meterbase-tls`
Above names are hard-coded.


v2/controllers/marketplace/certissuer_controller.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	for _, cm := range configMapList.Items {
+		if _, ok := cm.Annotations["service.beta.openshift.io/inject-cabundle"]; ok {
+			if _, ok := cm.Data["service-ca.crt"]; ok {
+				err := r.InjectCACertIntoConfigMap(&cm)
+				if err != nil {
+					log.Error(err, "failed to inject CA certificate")
+				}
+			} else {
+				klog.V(1).Info("service-ca.crt field was not found in a configmap labeled with inject-cabundle")
```
zach-source, line 86: Use of klog here. Please use the logger on the object. `reqlogger`

v2/controllers/marketplace/certissuer_controller.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *CertIssuerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if r.cfg.Infrastructure.HasOpenshift() {
+		return nil
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&marketplaceredhatcomv1beta1.CertIssuer{}).
```
zach-source, line 172: Not using CertIssuer here. You're really watching for the configmaps. We can remove the CRD and just watch for those configmaps/secrets.

v2/pkg/utils/certificates/cert_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"reflect"
+
+	"github.com/cloudflare/cfssl/helpers"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/types"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var _ = Describe("CertIssuer", func() {
```
zach-source, line 30: Good tests!

v2/pkg/inject/injectable.go
```diff
@@ -113,11 +115,26 @@ type KubeInterface interface {
 	InjectKubeInterface(kubernetes.Interface) error
 }
 
+type CertIssuer interface {
```
zach-source, line 118: Nice injector use.

zach-source (changes requested):
@MichalWrobel3 Overall looks real good, the biggest problem though is the inclusion of a new CRD that you're not using. We can remove that CRD and all the files related to it. Instead write your controller to look for the resources you care about (what you're listing). We can talk more about that if you need help.

zach-source:
This work is outdated and a new approach would need to occur.

PR #153, opened 2021-04-06
Upload/Download implementation for Airgap services

rajat-godi:
Changes include:
- GRPC API implementation for Upload/Download
- Download implementation for CLI

**Note:**
For some of the files under the database and server package, unit tests have been written under a separate package.
The reason for that is the functions under those packages are exported thus it allows us to test any issues while consuming those functions in external packages. Additionally it avoids import cycles in tests.


airgap/v2/cmd/client/download/download.go
```diff
@@ -0,0 +1,193 @@
...
+		} else {
+			bs = append(bs, data...)
+		}
+	}
+
+	outFile, err := os.Create(dc.outputDirectory + string(os.PathSeparator) + name)
+	if err != nil {
+		return fmt.Errorf("error while creating output file: %v", err)
+	}
+
```
zach-source, line 182:
```suggestion
defer outFile.Close()
```

We should defer just in case we error on write. Since this is a client it probably won't matter but it's a good practice.

airgap/v2/cmd/client/download/download_test.go
```diff
@@ -0,0 +1,221 @@
...
+			errMsg: "",
+		},
+		{
+			name:   "invalid download request with no file name/id provided",
+			dc:     &DownloadConfig{},
+			errMsg: "file id/name is blank",
+		},
+	}
+
+	for _, tt := range tests {
```
zach-source, line 192: Nice tests!

airgap/v2/cmd/client/download/download_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	res, err := clientStream.CloseAndRecv()
+	if err != nil {
+		t.Fatalf("error: during stream close and recieve: %v", err)
+	}
+	t.Log(fmt.Sprintf("Received response: %v", res))
+
+	downloadClient := fileretreiver.NewFileRetreiverClient(conn)
+	od, _ := os.Getwd()
+	tests := []struct {
```
zach-source, line 168: Like the use of a struct and table approach.

airgap/v2/pkg/dqlite/setup.go
```diff
@@ -0,0 +1,138 @@
...
+func (dc *DatabaseConfig) TryMigrate(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	defer cancel()
+	cli, err := dc.app.Leader(ctx)
+	if err != nil {
+		dc.Log.Error(err, "Could not find leader for migration")
+		return err
+	}
+
+	dc.Log.Info("Verifying leadership before migration")
```
zach-source, line 110: Great job here.

airgap/v2/pkg/dqlite/setup.go
```diff
@@ -0,0 +1,138 @@
...
+		return err
+	}
+
+	dc.app = app
+	dc.dqliteDB = conn
+	return conn.Ping()
+}
+
+// TryMigrate ensures that only the leader performs database migration
+func (dc *DatabaseConfig) TryMigrate(ctx context.Context) error {
```
zach-source, line 101: Good name of the function.

airgap/v2/pkg/dqlite/driver/dqlite.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	for k, v := range wrapper.dialector.ClauseBuilders() {
+		db.ClauseBuilders[k] = v
+	}
+	return
+}
+
+func (wrapper Dialector) ClauseBuilders() map[string]clause.ClauseBuilder {
+	return map[string]clause.ClauseBuilder{
```
zach-source, line 71: You can call the dialector on the struct.
```suggestion
	return wrapper.dialector.ClauseBuidlers()
```

zach-source: Please update the other functions in this driver to use the underlying dialector on the struct. That's why it's a wrapper.

airgap/v2/pkg/database/database.go
```diff
@@ -0,0 +1,105 @@
...
+func (d *Database) DownloadFile(finfo *v1.FileID) (*models.Metadata, error) {
+
+	var meta models.Metadata
+
+	fileid := strings.TrimSpace(finfo.GetId())
+	filename := strings.TrimSpace(finfo.GetName())
+
+	// Perform validations and query
+	if len(fileid) != 0 {
+		d.DB.Where("provided_id = ?", fileid).Order("created_at desc").Preload(clause.Associations).First(&meta)
```
zach-source, line 93: Let's format this a bit better. Couple places where you can use this to make it more readable
```suggestion
		d.DB.Where("provided_id = ?", fileid)
		  .Order("created_at desc")
		  .Preload(clause.Associations)
		  .First(&meta)
```

airgap/v2/pkg/database/database.go
```diff
@@ -0,0 +1,105 @@
...
+	} else if len(filename) != 0 {
+		d.DB.Where("provided_name = ?", filename).Order("created_at desc").Preload(clause.Associations).First(&meta)
+	} else {
+		return nil, fmt.Errorf("file id/name is blank")
+	}
+
+	if reflect.DeepEqual(meta, models.Metadata{}) {
+		return nil, fmt.Errorf("no file found for provided_name: %v / provided_id: %v", filename, fileid)
+	}
+	d.Log.Info(fmt.Sprintf("Retreived file of size: %v with id: %v", meta.Size, meta.FileID))
```
zach-source, line 103:
```suggestion
	d.Log.Info("Retreived file of size", "size", meta.Size, "id", meta.FileID))
```

logr.Logger interface should allow this, it's a little cleaner and can be searched in log output easier.

zach-source (changes requested):
Overall great job - few things we want to change.

* Your dqlite driver wrapper is reimplementing code when it should just call the driver on the struct.
* Some of your formatting is a bit off. Make sure you run `go fmt` on all your files and look to break out those long gorm calls with new lines.
* Review your log calls, you don't need to format them. You can use logr.Logger's functions var args to have a set of key to values.

zach-source (approved):

PR #154, opened 2021-04-09
Feature/monitoring stack

dacleyra:
Tests
- Add the Health conditions to the e2e test asserts (debt from another pr)
- Add UserWorkloadMonioting conditions to e2e test asserts
- Add userworkloadmonitoring-test e2e test set, described in tests/v2/e2e/userworkloadmonitoring-test/userworkloadmonitoring.md
- Modify meterbase_controller_test.go so it can discern RHOS version and handle lack of rhm-meterbase-additional-scrape-configs on 4.6

API Spec
- Add UserWorkloadMonitoringEnabled flag to MeterBase spec, add condition type,reason,messages for tracking status

Assets
- Add kube-state-metrics, kubelet, rhm-metric-state ServiceMonitors for UWM/4.6 prom instead of additional scrape configs
- Add a Secret that will be populated with redhat-marketplace-operator, such that there is a static named token that can be referenced for prometheus connection
- Add a prometheus deployment for 4.6 (prom v0.42.1). This is equivalent to UserWorkloadMonitoring and allows using ServiceMonitors with bearerTokenSecret references instead of hardcoded files
- Rename metric-state assets to distiguish 4.5 vs. 4.6

Controllers
- MeterBase 
  - Watches for changes to UserWorkloadMonioting ConfigMap assets
	- Handles UserWorkloadMonitoring and tracks the state in the status, including a 1d transitional
	- Installation of ServiceMonitors
	- Seperate out PrometheusServingCertsCABundle since it is also used by metric-state, in the UserWorkloadMonitoring case

- MeterDefinition
  - Relies on MeterBase instnace to determine which prometeus to query
	- Remove reconcile.Request from health funcs, it was never used

Configuration & assets
  - Handle using prom 4.6 or 4.5 images
	- Add semver parsing to infrastructure for version comparisons
  - Handle the new ServiceMonitors and versioned deployments in the factory
	- prevent a panic in getCredFromSecret if map access cameback nil
	
Prometheus
  - handle UserWorkloadMonitoring for prom api, client
	
Finalizer bug
  - With how finalizer result was being used, it should return Finished when successfully done, so the reconciler stops. Prior it would continue to install subcomponents, even after finalizer uninstalled things. Set to to consider Fininshed if the prior actions were success/Continue or already-deleted/OnNotFound. Deletes should be a lot snappier, since it is not churning now.


v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -265,6 +297,68 @@ func (r *MeterBaseReconciler) Reconcile(request reconcile.Request) (reconcile.Re
...
+			instance,
+			userWorkloadMonitoringEnabledSpec,
+			userWorkloadMonitoringEnabledOnCluster,
+		); result.Is(Error) || result.Is(Requeue) {
+			if err != nil {
+				return result.ReturnWithError(merrors.Wrap(err, "error updating status"))
+			}
+			return result.Return()
+		}
+	} else if condition := instance.Status.Conditions.GetCondition(marketplacev1alpha1.ConditionUserWorkloadMonitoringEnabled); condition != nil {
```
zach-source, line 320:
```suggestion
	} 
	
	userWorkloadCondition := instance.Status.Conditions.GetCondition(marketplacev1alpha1.ConditionUserWorkloadMonitoringEnabled)
```

Your first if statement if true should requeue regardless.

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -265,6 +297,68 @@ func (r *MeterBaseReconciler) Reconcile(request reconcile.Request) (reconcile.Re
...
+	userWorkloadMonitoringEnabledSpec := instance.Spec.UserWorkloadMonitoringEnabled
+	userWorkloadMonitoringEnabledOnCluster, result := isUserWorkloadMonitoringEnabledOnCluster(cc, r.cfg.Infrastructure, reqLogger)
+	if !result.Is(Continue) {
+		return result.Return()
+	}
+	userWorkloadMonitoringEnabled := userWorkloadMonitoringEnabledOnCluster && userWorkloadMonitoringEnabledSpec
+
+	if instance.Status.Conditions.IsUnknownFor(marketplacev1alpha1.ConditionUserWorkloadMonitoringEnabled) {
+		// Set initial UWM status
+		if result, err := updateUserWorkloadMonitoringEnabledStatus(cc,
```
zach-source, line 310: Don't need an if here. We should requeue regardless. Only if condition should be if err after result is returned.

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -265,6 +297,68 @@ func (r *MeterBaseReconciler) Reconcile(request reconcile.Request) (reconcile.Re
...
+			userWorkloadMonitoringEnabledSpec,
+			userWorkloadMonitoringEnabledOnCluster,
+		); result.Is(Error) || result.Is(Requeue) {
+			if err != nil {
+				return result.ReturnWithError(merrors.Wrap(err, "error updating status"))
+			}
+			return result.Return()
+		}
+	} else if condition := instance.Status.Conditions.GetCondition(marketplacev1alpha1.ConditionUserWorkloadMonitoringEnabled); condition != nil {
+		if userWorkloadMonitoringEnabled != instance.Status.Conditions.IsTrueFor(marketplacev1alpha1.ConditionUserWorkloadMonitoringEnabled) {
```
zach-source, line 321:
```suggestion
		if userWorkloadCondition.IsTrue() {
```

zach-source: ... And other changes below

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -265,6 +297,68 @@ func (r *MeterBaseReconciler) Reconcile(request reconcile.Request) (reconcile.Re
...
+					return result.ReturnWithError(merrors.Wrap(err, "error updating status"))
+				}
+				return result.Return()
+			}
+		}
+	}
+
+	// Determine state of UWM transition
+	transitionTimeExpired := false
+	if condition := instance.Status.Conditions.GetCondition(marketplacev1alpha1.ConditionUserWorkloadMonitoringEnabled); condition != nil {
```
zach-source, line 353: User earlier value from setting var

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -285,31 +379,81 @@ func (r *MeterBaseReconciler) Reconcile(request reconcile.Request) (reconcile.Re
...
+
+			reqLogger.Info("returing result", "result", *result)
+			return result.Return()
+		}
+
+		promStsNamespacedName = types.NamespacedName{
+			Namespace: utils.OPENSHIFT_USER_WORKLOAD_MONITORING_NAMESPACE,
+			Name:      utils.OPENSHIFT_USER_WORKLOAD_MONITORING_STATEFULSET_NAME,
+		}
+	} else {
```
zach-source, line 418:
```suggestion
	} else { // if userWorkload is not enabled
```

zach-source: Try to comment your else with the condition you are assuming in them. Helps us spot errors and remembering what code is doing.

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -646,9 +812,20 @@ func (r *MeterBaseReconciler) reconcilePrometheusOperator(
 		Patcher: r.patcher,
 	}
 
-	watchNamespace, _ := getWatchNamespace()
```
zach-source, line 649: Why do we not use the getWatchNamespace function? Think this may break our operator deployment.

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -681,14 +861,17 @@ func (r *MeterBaseReconciler) installMetricStateDeployment(
 ) []ClientAction {
 	deployment := &appsv1.Deployment{}
 	service := &corev1.Service{}
-	serviceMonitor := &monitoringv1.ServiceMonitor{}
+	serviceMonitor0 := &monitoringv1.ServiceMonitor{}
```
zach-source, line 864: Can we enhance this name a bit? serviceMonitorShortName i.e. like serviceMonitorMetricState

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -704,13 +887,40 @@ func (r *MeterBaseReconciler) installMetricStateDeployment(
...
-			serviceMonitor,
+			serviceMonitor2,
 			func() (runtime.Object, error) {
 				return factory.MetricStateServiceMonitor()
 			},
 			args,
 		),
 	}
+
+	if r.cfg.Infrastructure.OpenshiftParsedVersion().GTE(utils.ParsedVersion460) {
```
zach-source, line 898: let's move serviceMonitor0 and serviceMonitor var def here so we know it's apart of this functionality.

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -1121,12 +1337,24 @@ func (r *MeterBaseReconciler) uninstallMetricState(
 ) []ClientAction {
 	deployment, _ := factory.MetricStateDeployment()
 	service, _ := factory.MetricStateService()
-	sm, _ := factory.MetricStateServiceMonitor()
+	sm0, _ := factory.MetricStateServiceMonitor()
```
zach-source, line 1340: sm0 is fine here for the delete, small enough function to know what we're doing.

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -1167,24 +1386,42 @@ func (r *MeterBaseReconciler) uninstallPrometheus(
...
-		HandleResult(
-			GetAction(types.NamespacedName{Namespace: deployment.Namespace, Name: deployment.Name}, deployment),
-			OnContinue(DeleteAction(deployment))),
 		HandleResult(
 			GetAction(types.NamespacedName{Namespace: prom.Namespace, Name: prom.Name}, prom),
 			OnContinue(DeleteAction(prom))),
 	)
 }
 
+func (r *MeterBaseReconciler) installPrometheusServingCertsCABundle(
```
zach-source, line 1398: factory is apart of the MeterBaseReconciler - we can refactor these functions to use that. The original code passed it in functions but if its' in a struct we can clean this function up a bit.

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -1316,11 +1553,115 @@ func labelsForPrometheusOperator(name string) map[string]string {
...
+}
+
+func updateUserWorkloadMonitoringEnabledStatus(
+	cc ClientCommandRunner,
+	instance *marketplacev1alpha1.MeterBase,
+	userWorkloadMonitoringEnabledSpec bool,
+	userWorkloadMonitoringEnabledOnCluster bool,
+) (*ExecResult, error) {
+
+	if userWorkloadMonitoringEnabledSpec && userWorkloadMonitoringEnabledOnCluster {
```
zach-source, line 1633: Just a nit. More a style thing but you can write this without else since you are returning in every function.

v2/pkg/config/config.go
```diff
@@ -150,6 +151,29 @@ func ProvideInfrastructureAwareConfig(
 			return nil, errors.Wrap(err, "failed to parse config")
 		}
 
+		// Use v4.6 images on Openshift 4.6 instead of default v4.5 images
+		// But otherwise respect an override
+		if inf.OpenshiftParsedVersion().GTE(utils.ParsedVersion460) {
+			if cfg.RelatedImages.Prometheus == "registry.redhat.io/openshift4/ose-prometheus:v4.5" {
```
zach-source, line 157: This is a little wordy but I think it's fine for now. Sometimes better to be explicit.

zach-source (changes requested):
Some code cleanup but otherwise great PR.

v2/controllers/marketplace/meterdefinition_controller.go
```diff
@@ -176,7 +178,44 @@ func (r *MeterDefinitionReconciler) Reconcile(request reconcile.Request) (reconc
 		reqLogger.Error(result.GetError(), "Failed to update status.")
 	}
 
-	isReporting, err := r.verifyReporting(cc, instance, request, reqLogger)
+	// Fetch the MeterBase instance
+	meterbase := &v1alpha1.MeterBase{}
+	result, _ = cc.Do(context.TODO(),
+		GetAction(types.NamespacedName{
+			Name:      utils.METERBASE_NAME,
+			Namespace: "openshift-redhat-marketplace",
```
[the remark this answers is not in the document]
dacleyra: Namespace is now from r.cfg.DeployedNamespace

v2/controllers/marketplace/meterdefinition_controller.go
```diff
@@ -259,22 +301,133 @@ func (r *MeterDefinitionReconciler) addFinalizer(instance *v1beta1.MeterDefiniti
...
+		return queryPreviewResult, err
 	}
 
 	reqLogger.Info("generatring meterdef preview")
-	queryPreviewResult, err := generateQueryPreview(instance, prometheusAPI, reqLogger)
+	return generateQueryPreview(instance, prometheusAPI, reqLogger)
+}
+
+/*
+func (r *MeterDefinitionReconciler) queryPreviewOLD(cc ClientCommandRunner, instance *v1beta1.MeterDefinition, request reconcile.Request, reqLogger logr.Logger, userWorkloadMonitoringEnabled bool) ([]common.Result, error) {
```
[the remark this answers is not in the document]
dacleyra: Commented functions are removed

v2/controllers/marketplace/meterbase_controller.go
```diff
@@ -129,6 +135,27 @@ func (r *MeterBaseReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return (e.MetaNew.GetName() == utils.OPENSHIFT_CLUSTER_MONITORING_CONFIGMAP_NAME && e.MetaNew.GetNamespace() == utils.OPENSHIFT_MONITORING_NAMESPACE) ||
+				(e.MetaNew.GetName() == utils.OPENSHIFT_USER_WORKLOAD_MONITORING_CONFIGMAP_NAME && e.MetaNew.GetNamespace() == utils.OPENSHIFT_USER_WORKLOAD_MONITORING_NAMESPACE) ||
+				(e.MetaNew.GetName() == utils.KUBELET_SERVING_CA_BUNDLE_NAME && e.MetaNew.GetNamespace() == utils.OPENSHIFT_MONITORING_NAMESPACE)
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return (e.Meta.GetName() == utils.OPENSHIFT_CLUSTER_MONITORING_CONFIGMAP_NAME && e.Meta.GetNamespace() == utils.OPENSHIFT_MONITORING_NAMESPACE) ||
+				(e.Meta.GetName() == utils.OPENSHIFT_USER_WORKLOAD_MONITORING_CONFIGMAP_NAME && e.Meta.GetNamespace() == utils.OPENSHIFT_USER_WORKLOAD_MONITORING_NAMESPACE) ||
+				(e.Meta.GetName() == utils.KUBELET_SERVING_CA_BUNDLE_NAME && e.Meta.GetNamespace() == utils.OPENSHIFT_MONITORING_NAMESPACE)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
```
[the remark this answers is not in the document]
dacleyra: move to fn isOpenshiftMonitoringObj

v2/controllers/marketplace/meterbase_controller.go
dacleyra: This is intentional. Do not want to updateUserWorkloadMonitoringEnabledStatus if the state is Transitioning, until the transition timer has expired.

v2/controllers/marketplace/meterbase_controller.go
dacleyra: GetCondition can return nil, not prudent to check since accessing condition.Reason?

v2/controllers/marketplace/meterbase_controller.go
dacleyra: ok, moved

v2/controllers/marketplace/meterbase_controller.go
dacleyra: ok

v2/controllers/marketplace/meterbase_controller.go
dacleyra: ok

v2/controllers/marketplace/meterbase_controller.go
dacleyra: watchNamespace was used to feed the factory for NewPrometheusOperatorDeployment
NewPrometheusOperatorDeployment was changed
the nsValues are now the deny-namespaces list, that is, the platform openshift- namespaces (openshift.io/cluster-monitoring0

v2/controllers/marketplace/meterbase_controller.go
dacleyra: changed to be in step with the factory name

v2/controllers/marketplace/meterbase_controller.go
dacleyra: updated this func, as well as the others to use r.factory

v2/controllers/marketplace/meterbase_controller.go
dacleyra: I don't think so... this needs to be a comparison between userWorkloadMonitoringEnabled  (the spec && cluster setting) and the state in the status, not just an if status

zach-source (approved):

zach-source (approved):

PR #155, opened 2021-04-20
feature/reportGroupByHandling

zach-source:
- Adds functionality and fields to the Meter Report to build reports where the data reported is always captured.
- Add additional fields to MeterDefinition (label and unit). These fields are templatable.
- Creates a schema version and rework of the meter report format.


mxpaspa (approved):
LGTM.

PR #168, opened 2021-05-19
hackathon work to create meterdefinition key value store

mxpaspa:


mxpaspa:
see 172 for current work

PR #172, opened 2021-05-22
Hackathon work to install a CSV's associated meterdefinition

mxpaspa:


mxpaspa:
see #177 for current work

PR #176, opened 2021-06-02
Feature/airgap deployment

dacleyra:
Add the airgap target to the base level Makefile

Add liblz4 to build, as it was failing without
Adjust copy dir based on build loc, this may need to be streamlined better

Refactor airgap Makefile to be in line with other components, this may need to be streamlined better

Add an example secure client config yaml

Add a token flag for the airgap client, in order to pass in a service account token for auth via kube-rbac-proxy

~~Use a fork of go-dqlite which allows for the use of FQDN hostnames as node endpoints, instead of only ip addresses
Statefulset has stable FQDN, but not a stable IP in the case of a pod delete~~
Once TLS is enabled, the FQDN is used for the cluster configuration and directory, so this ended up being unnecessary
db cluster traffic is secured with TLS on pod-to-pod traffic port 9001

Add --allowerasing --nobest for base.Dockerfile ubi image, as it was failing on latest ubi release

add airgap targets to v2/Makefile

Add flag to MeterBase for enabling airgap dataservice

Add Route for external access to dataservice

Administrator must enable http2 for ingress router
oc annotate ingresses.config/cluster ingress.operator.openshift.io/default-enable-http2=true

https2/grpc -> Router
-> rhm-dqlite:8002 Service https2 kube-rbac-proxy sidecar
-> localhost:8001 http2 go-dqlite api 

Add dqlite image override envvar option

Add rbac for controller to use Routes (all the things)
Add rbac for grpc method (nonResourceUrl create)

Add dataservice_controller

Add Route to scheme & controller to manager

Add factory items for dataservice
Add update funcs to enforce spec


airgap/v2/Makefile
```diff
@@ -10,7 +80,7 @@ build:
 demo: build
 	docker-compose up
 
-generate: protoc
+generate-old: protoc
```
zach-source, line 83:
```suggestion
generate-proto: protoc
```

zach-source (changes requested):

zach-source (approved):

PR #180, opened 2021-06-10
Feature/use embed fs

dacleyra:
Move up to go 1.16
Use embed instead of go-bindata


v2/pkg/manifests/factory.go
```diff
@@ -68,8 +67,19 @@ const (
...
+	bts, err := assets.ReadFile(filename)
+	if err != nil {
+		panic(err)
+	}
+	return bts
+}
+
 func MustAssetReader(asset string) io.Reader {
-	return bytes.NewReader(MustAsset(asset))
+	return bytes.NewReader(MustReadFileAsset(asset))
```
zach-source, line 82: Perfect!

zach-source (changes requested):

zach-source (dismissed):

PR #183, opened 2021-06-15
razee cc cleanup

dacleyra:
WIP

- Change assets to be defined primarily by yaml
- Use controllerutil.CreateOrUpdate instead of cc.Do pattern
- CreateOrUpdate takes advantage of the fact that yaml Decode will merge fields, rather than reset fields in the struct.
- Therefore on the update func(), the spec in the factory object yaml will be enforced, but it can otherwise be annotated or labeled by the end user.


zach-source (approved):
Tested, looks good to me.

PR #210, opened 2021-07-16
updated marketplaceconfig desc, bumped images to 2.3.0

ab-ibm:


dacleyra (commented):
a few grammar & punctuation updates needed

I believe v2/version/version.go:const Version = "2.2.3"
needs to be updated to 2.3.0

PR #212, opened 2021-07-20
file-server-tls merge with develop

mxpaspa:
combines work from `feature/file-server-tls` and a merge with latest develop


mxpaspa:
see #214 for latest work

PR #216, opened 2021-07-27
added version

ab-ibm:


v2/bundle/manifests/redhat-marketplace-operator.clusterserviceversion.yaml
```diff
@@ -1172,7 +1172,8 @@ spec:
       targetPort: 9443
       type: MutatingAdmissionWebhook
       webhookPath: /mutate-marketplace-redhat-com-v1beta1-meterdefinition
-    - admissionReviewVersions: null
+    - admissionReviewVersions:
```
zach-source, line 1175: This code is generated - I'm pretty sure this is going to be overwritten. This is probably one the webhook annotations found in this file v2/apis/marketplace/v1beta1/meterdefinition_webhook.go

zach-source (changes requested):

zach-source:
Bug branches should be named bugfix/ - this will auto run builds and checks. @ab-ibm

PR #219, opened 2021-07-29
bugfix: fixing CVE-2020-26160

zach-source:
This removes the github.com/dgrijalva/jwt-go project for the replacement github.com/go-jwt/jwt and fixes the CVE-2020-26160. Due to how go modules work - we have to use a fork of our own called github.com/redhat-marketplace/jwt that adds go modules. The old project is removed using the go mod replaces clause and is not present in go.sum files through the project.


mxpaspa (dismissed):
LGTM

PR #242, opened 2021-09-09
Feature/reporterv2schema

dacleyra:
- add flag "reporterSchema" to switch between v1alpha1 and v2alpha1 schema versions, using the reporterv1 or reporterv2 accordingly
- Add the schemav2alpha1 structs and builder
- Add the MarketplaceReporterV2. Primary difference from MarketplaceReporter (v1) is just for its Process() and WriteReport() funcs to work with schemav2alpha1
- reporterv2 test copies the same two tests as reporterv1 and checks the new report output format. A 3rd test is added to call both the reporterv1 and reporterv2 and validate the v1/metricId = v2/eventId
- builder test checks that the sample report files unmarshal to the new data struct correctly
- add a cos-s3 uploader
- To change schema and enable the uploader
  - `oc create secret generic rhm-cos-uploader-secret --from-file=config.yaml`
  - set the envs on the redhat-marketplace-controller-manager Deployment
```
        env:
        - name: REPORTERSCHEMA
          value: v2alpha1
        - name: UPLOADTARGETSOVERRIDE
          value: cos-s3
```


zach-source (approved):
LGTM - we'll refactor some of the implementation details in future PR but bones of it is great.

PR #248, opened 2021-09-14
validate user-workload-monitoring-config

mxpaspa:
-  checks for `user-workload-monitoring-config` on cluster
if found: 
- validates that retention time is at least 168h
- validates that disk space is at least 40Gi
- updates meterbase status conditions with validation errors

the config below should pass validation. You can test by changing retention and storage and ensuring the correct conditions get set
`user-workload-monitoring-config`: 
```YAML
apiVersion: v1
kind: ConfigMap
metadata:
  name: user-workload-monitoring-config
  namespace: openshift-user-workload-monitoring
data:
  config.yaml: |
    prometheus: 
      retention: 168h 
      volumeClaimTemplate:
        spec:
          resources:
            requests:
              storage: 40Gi
```


zach-source (approved):
LG(reat)TM

PR #258, opened 2021-09-30
bugfix: Configurable Resources

zach-source:
- Adds a env flag called `CONFIG` where you can set container resources
- Adds function to factory to match containers by name and override container resources based on the `CONFIG` configuration.

For example, setting a config env flag on the subscription override the limits for "system" and "system2" container names.
```
CONFIG: {"resources":{"containers":{"system":{"limits":{"cpu":"500m"}},"system2":{"limits":{"cpu":"1"}}}}}
```

### DIFF
--- a/v2/pkg/config/config.go
+++ b/v2/pkg/config/config.go
@@ -15,6 +15,8 @@
 package config
 
 import (
+	"bytes"
+	"reflect"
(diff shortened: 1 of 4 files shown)


mxpaspa (approved):
LGTM

PR #260, opened 2021-10-05
send events for disabled insights

dacleyra:
create broadcaster at top level that can optionally be passed to Task functions
task records ReportJobError type to events on the Job object before returning
the Event will send the ReportJobError.ErrorMessage to the Event (not the stack)
Is there anything more specific you want in the Event message?

ReconcileTask.Run was accumulating errors without returning them
Which I modified for CanRunUploadTask, but not range meterReportsToRun
Do we want to do something specific about this?

### DIFF
--- a/reporter/v2/cmd/reporter/reconciler/reconcile.go
+++ b/reporter/v2/cmd/reporter/reconciler/reconcile.go
@@ -16,13 +16,19 @@ package reconciler
 
 import (
 	"context"
-	"errors"
 	"os"
 	"time"
+
+	"emperror.dev/errors"
 
 	"github.com/gotidy/ptr"
 	"github.com/redhat-marketplace/redhat-marketplace-operator/reporter/v2/pkg/reporter"
 	"github.com/spf13/cobra"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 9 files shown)


reporter/v2/cmd/reporter/reconciler/reconcile.go
```diff
@@ -81,23 +82,33 @@ var ReconcileCmd = &cobra.Command{
 		}
 		cfg.SetDefaults()
 
+		broadcaster, stopBroadcast, err := reporter.NewEventBroadcaster(ctx, cfg)
+		if err != nil {
+			log.Error(err, "couldn't initialize event broadcaster")
+			os.Exit(1)
+		}
```
zach-source, line 89:
```suggestion
		}
		defer stopBroadcast()
```

reporter/v2/cmd/reporter/reconciler/reconcile.go
```diff
@@ -81,23 +82,33 @@ var ReconcileCmd = &cobra.Command{
...
 			ctx,
 			cfg,
+			broadcaster,
 			reporter.Namespace(namespace),
 		)
 
 		if err != nil {
 			log.Error(err, "couldn't initialize task")
+			stopBroadcast()
 			os.Exit(1)
```
zach-source, line 101: We can make this a little clearer and easier for you to handle. Let's move stopBroadcast to the defer. We can change the command function to return an error `RunE`, then just return the errors here instead of os.Exit(1). Returning an error will cause it to exit so we don't need to call os.Exit any longer. Then we can remove the stopBroadcast() from the if's and end, it'll run when the function exits.

reporter/v2/pkg/reporter/task.go
```diff
@@ -329,6 +334,26 @@ func getMeterDefinitionReferences(
...
+// https://github.com/kubernetes/kubernetes/issues/94906
+func provideReporterEventBroadcaster(kubeclientset kubernetes.Interface) (record.EventBroadcaster, func()) {
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartStructuredLogging(0)
+	shutdownInterface := eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeclientset.CoreV1().Events("")})
+	return eventBroadcaster, func() {
+		shutdownInterface.Stop()
+		logger.Info("Wait for event broadcaster to stop...")
+		time.Sleep(10 * time.Second)
+	}
```
zach-source, line 349: Nice use of a callback. There is a kubernetes Closer concept but it's hard to implement, function works fine.

zach-source (changes requested):
Small changes on task otherwise looks great.

zach-source (approved):

PR #261, opened 2021-10-05
uwm event for DefaultStorageClassNotFound

dacleyra:
- Send an event for DefaultClassNotFound for userWorkloadMonitoring, just as we do with rhm-prometheus setup.
- Send the event, but do not stop reconciliation, as user could have setup a storageclass manually for UWM


zach-source:
@dacleyra Can you update this to target 2.4.1? I've changed the base but looks like it has some conflicts.

PR #278, opened 2021-10-21
update remoteresources3 and watch-keeper images

ab-ibm:


zach-source:
We can test this with the 2.5.0 releases.

PR #284, opened 2021-11-03
WIP Feature/cert pipeline

dacleyra:
To pass certification checks
- annotations.yaml needs
  - openshift.versions
	- CHANNELS, DEFAULT_CHANNELS with default; must be set to generate their required associated annotations
- remove the operator service account yaml, this should be created by OLM & fails a cert check
  - didn't seem to be a great way to remove it via a kustomize option; make bundle vs. make install needing the sa

### DIFF
--- a/v2/hack/certify/certify.sh
+++ b/v2/hack/certify/certify.sh
@@ -6,6 +6,9 @@ SLEEP_SHORT="${SLEEP_SHORT:-2}"
 CERT_NAMESPACE="rhm-certification"
 KUBECONFIG=$HOME/.kube/config
 VERSION="2.4.0"
+IMAGE_REGISTRY=quay.io/dacleyra
+CHANNELS=beta,stable
+DEFAULT_CHANNEL=stable
 
 # Check Subscriptions: subscription-name, namespace
 checksub () {
@@ -117,6 +120,8 @@ oc apply -R -f ansible/roles/operator-pipeline/templates/openshift/pipelines
 oc apply -R -f ansible/roles/operator-pipeline/templates/openshift/tasks
 cd $CWD
 
+oc apply -f https://raw.githubusercontent.com/tonytcampbell/operator-pipelines/preflight-fixes/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight.yml
+
 
 # Generate the bundle and add it to the fork
 cd $TMP_DIR
@@ -136,12 +141,6 @@ rm -Rf  $TMP_DIR/certified-operators-preprod/operators/redhat-marketplace-operat
 
 echo "organization: redhat-marketplace" > $TMP_DIR/certified-operators-preprod/config.yaml
 echo "cert_project_id: 5f68c9457115dbd1183ccab6" > $TMP_DIR/certified-operators-preprod/operators/redhat-marketplace-operator/ci.yaml
-
-cd $TMP_DIR
-curl -L https://github.com/mikefarah/yq/releases/download/v4.13.5/yq_linux_amd64.tar.gz | tar -xz 
-./yq_linux_amd64 eval '
-  .annotations."com.redhat.openshift.versions" = "v4.6-v4.9"
-' -i $TMP_DIR/certified-operators-preprod/operators/redhat-marketplace-operator/$VERSION/metadata/annotations.yaml
 
 cd $TMP_DIR/certified-operators-preprod
 git add --all
--- a/v2/hack/certify/certify.sh
+++ b/v2/hack/certify/certify.sh
@@ -125,7 +125,7 @@ echo "cert_project_id: 5ec3fc8628834587a6b85c2a" > $TMP_DIR/certified-operators-
 cd $TMP_DIR
 curl -L https://github.com/mikefarah/yq/releases/download/v4.13.5/yq_linux_amd64.tar.gz | tar -xz 
 ./yq_linux_amd64 eval '
-  .metadata.annotations."com.redhat.openshift.versions" = "v4.6-v4.9"
+  .annotations."com.redhat.openshift.versions" = "v4.6-v4.9"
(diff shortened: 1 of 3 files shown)


v2/hack/certify/certify.sh
```diff
@@ -0,0 +1,164 @@
...
+# kustomize questionable capability to remove the service account
+rm -Rf  $TMP_DIR/certified-operators-preprod/operators/redhat-marketplace-operator/$VERSION/manifests/redhat-marketplace-operator_v1_serviceaccount.yaml
+
+echo "organization: redhat-marketplace" > $TMP_DIR/certified-operators-preprod/config.yaml
+echo "cert_project_id: 5f68c9457115dbd1183ccab6" > $TMP_DIR/certified-operators-preprod/operators/redhat-marketplace-operator/ci.yaml
+
+cd $TMP_DIR/certified-operators-preprod
+git add --all
+git commit -m $VERSION
+git push
```
zach-source, line 146: Can you use branches or it has to be a main commit&push?

v2/hack/certify/certify.sh
```diff
@@ -0,0 +1,164 @@
...
+	unset PHASE
+	until [ "$PHASE" == "Complete" ]
+	do
+		PHASE=$(oc get installplan $INSTALL_PLAN -n $2 --output=jsonpath={.status.phase})
+    if [ "$PHASE" == "Failed" ]; then
+      set +x
+      sleep 3
+      echo "InstallPlan $INSTALL_PLAN for subscription $1 failed."
+      echo "To investigate the reason of the InstallPlan failure run:"
+      echo "oc describe installplan $INSTALL_PLAN -n $2"
```
zach-source, line 40: Nice comment!

v2/hack/certify/certify.sh
```diff
@@ -0,0 +1,164 @@
...
+cd $TMP_DIR/certified-operators-preprod
+git add --all
+git commit -m $VERSION
+git push
+
+
+# Run the Pipeline
+
+cd $TMP_DIR/operator-pipelines
+curl https://mirror.openshift.com/pub/openshift-v4/clients/pipeline/0.19.1/tkn-linux-amd64-0.19.1.tar.gz | tar -xz
```
zach-source, line 152: Add a check for a local tkn or one on the path. Prefer local if it's there. Print version.

v2/hack/certify/certify.sh
```diff
@@ -0,0 +1,164 @@
...
+
+# Run the Pipeline
+
+cd $TMP_DIR/operator-pipelines
+curl https://mirror.openshift.com/pub/openshift-v4/clients/pipeline/0.19.1/tkn-linux-amd64-0.19.1.tar.gz | tar -xz 
+
+GIT_REPO_URL=https://github.com/redhat-marketplace/certified-operators-preprod.git
+BUNDLE_PATH=operators/redhat-marketplace-operator/$VERSION
+
+./tkn pipeline start operator-ci-pipeline \
```
zach-source, line 157: Does the pipeline clean up after it's done?

v2/hack/certify/certify.sh
```diff
@@ -0,0 +1,164 @@
...
+  --param git_repo_url=$GIT_REPO_URL \
+  --param git_branch=stage \
+  --param bundle_path=$BUNDLE_PATH \
+  --param env=stage \
+  --workspace name=pipeline,volumeClaimTemplateFile=templates/workspace-template.yml \
```
zach-source, line 162: Does there need to exist a template/workspace-template.yml somewhere? Or is this in the other repo?

zach-source (changes requested):
Minor nits.

v2/hack/certify/certify.sh
dacleyra: they actually just changed the repo in the instructions, and the default branch was stage, now it is main
we can set this to our choice of branch, the git_branch is specified in the tekton run
let me know how you want to handle it

v2/hack/certify/certify.sh
dacleyra: checksub routine was written back in pre-ICPA days, before there was a product and ansible installer, and OLM was a little rougher around the edges

this at least got people to not immediately ask me blindly why a hacked up pre-prod install didn't work...

v2/hack/certify/certify.sh
dacleyra: pipeline will delete created artifacts whether the pipeline succeeded for failed
however, this is where we got into the failure related to the stuck finalizer problem, as a result of installing to our non default namespace

v2/hack/certify/certify.sh
zach-source: Let's create a branch named for the release off the version and pass that. It'll let us keep stuff separate. Maybe at some point we can trigger actions on merge. If that's too complicated to do right now we can do follow on. First priority is a release of 2.4.0

v2/hack/certify/certify.sh
dacleyra: this is all being done in a tmp directory, tkn won't exist run-to-run, unless you want it installed elsewhere

v2/hack/certify/certify.sh
dacleyra: At this point in the script, we are in the operator-pipelines dir, so the file is

https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/templates/workspace-template.yml

v2/hack/certify/certify.sh
zach-source: Ah, missed that. Carry on.

v2/hack/certify/certify.sh
zach-source: Hah - it's the little things tho.

zach-source (changes requested):
Still WIP

PR #285, opened 2021-11-04
disconnected flag

mxpaspa:


reporter/v2/pkg/reporter/reconcile_task.go
```diff
@@ -101,6 +101,67 @@ func (r *ReconcileTask) run(ctx context.Context) error {
 			logger.Error(err, "error uploading files from data service", details...)
 			return err
 		}
+	} else if !r.CanRunUploadTask(ctx) {
```
mxpaspa, line 104: @zach-source is this the right approach to update the report statuses ?

reporter/v2/pkg/reporter/reconcile_task.go
```diff
@@ -158,7 +219,7 @@ func (r *ReconcileTask) ReportTask(ctx context.Context, report *marketplacev1alp
 }
 
 func (r *ReconcileTask) CanRunUploadTask(ctx context.Context) bool {
```
zach-source, line 221: CanRunUploadTask should be true if you can do it. If you return just `IsDisconnected` then that would flip this. Try inversing the value.

reporter/v2/pkg/reporter/reconcile_task.go
```diff
@@ -158,7 +219,7 @@ func (r *ReconcileTask) ReportTask(ctx context.Context, report *marketplacev1alp
...
-	return true
+	return r.Config.IsDisconnected
```
zach-source, line 222:
```suggestion
	return !r.Config.IsDisconnected
```

reporter/v2/pkg/reporter/reconcile_task.go
zach-source:
```suggestion
	} else if r.CanRunUploadTask(ctx) {
```

reporter/v2/pkg/reporter/reconcile_task.go
```diff
@@ -101,6 +101,67 @@ func (r *ReconcileTask) run(ctx context.Context) error {
...
+			condition := marketplacev1alpha1.ReportConditionJobIsDisconnected
+			details := &marketplacev1alpha1.UploadDetails{}
+			details.Status = marketplacev1alpha1.UploadStatusNoop
+
+			if err := updateMeterReportStatus(ctx, r.K8SClient, metadata.ReportName, metadata.ReportNamespace,
+				func(m marketplacev1alpha1.MeterReportStatus) marketplacev1alpha1.MeterReportStatus {
+					m.Conditions.SetCondition(condition)
+
+					statuses = append(statuses, details)
+					m.UploadStatus.Append(statuses)
```
zach-source, line 150: Let's not update the UploadStatus. No need to put a row there. Conditions is fine.

reporter/v2/pkg/reporter/reconcile_task.go
```diff
@@ -101,6 +101,67 @@ func (r *ReconcileTask) run(ctx context.Context) error {
...
+
+					dataServiceStatus := m.UploadStatus.Get(uploaders.UploaderTargetDataService.Name())
+
+					if dataServiceStatus != nil {
+						m.DataServiceStatus = dataServiceStatus
+					}
+
+					return m
+				}); err != nil {
+				logger.Error(err, "failed to update meter report")
```
zach-source, line 160:
```suggestion
			condition := marketplacev1alpha1.ReportConditionJobIsDisconnected

			if err := updateMeterReportStatus(ctx, r.K8SClient, metadata.ReportName, metadata.ReportNamespace,
				func(m marketplacev1alpha1.MeterReportStatus) marketplacev1alpha1.MeterReportStatus {
					m.Conditions.SetCondition(condition)
					return m
				}); err != nil {
				logger.Error(err, "failed to update meter report")
```

zach-source (changes requested):

zach-source (approved):
Few small issues that came out with testing. I've added a commit to fix them.

PR #286, opened 2021-11-05
bugfix: fix rrs3 reconcilation

zach-source:
Changes
- Add unknown-fields attribute to RRS3 spec (required for v1 to work)
- Patches RRS3 deployment asset yaml to fix pathing issues with startup
- Patches razeedeployment controller to get past equality checks on secret data (infinite loop, yay)
- Removes HTTPS_PROXY being set to "" on RRS3 and watchkeeper.
- Resets permissions to 2.3.4
- Bumps limits on Prometheus Operator to prevent OOM


dacleyra (approved):
lgtm

PR #289, opened 2021-11-09
mergo for rrs3 CreateOrUpdate

dacleyra:
- Use mergo to merge/enforce spec on Deployment when using CreateOrUpdate
- mergo.WithOverride replaces slices, which could otherwise cause problems for things like volume list on upgrade


zach-source (approved):
Looks great!

PR #291, opened 2021-11-11
feature/dataServiceFiltering

zach-source:
- feature: adding filtering to dataservice
   - Adds support for using a list filter on the files request. 
   - Generic filter parser
   - GORM db layer to add filter to queries
- bugfix: remove extra field on report
   - fixes schema issue with v2alpha1 reports


dacleyra (approved):
lgtm

PR #306, opened 2022-01-26
requeue if olm.operatorNamespace isn't set yet

mxpaspa:
For: https://github.ibm.com/symposium/track-and-plan/issues/22915

Our predicates were looking for the csv namespace to match the namespace set by olm on `olm.operatorNamespace` upon creation of the csv
```

func csvFilter(metaNew metav1.Object) int {
	ann := metaNew.GetAnnotations()

	//annotation values
	ignoreVal, hasIgnoreTag := ann[ignoreTag]
	_, hasCopiedFrom := ann[olmCopiedFromTag]
	_, hasMeterDefinition := ann[utils.CSV_METERDEFINITION_ANNOTATION]

	sameNamespace := ann[olmNamespace] == metaNew.GetNamespace()

	switch {
	// has meterdef && is not an all namespace install
	case hasMeterDefinition && !hasCopiedFrom && sameNamespace:
		return 1
	// does not have a meterdef and
	case !hasMeterDefinition && (!hasIgnoreTag || ignoreVal != ignoreTagValue):
		return 2
	default:
	}

	return 0
}
```
The problem is, OLM is sometimes slow to set `olm.operatorNamespace` meaning our create predicate returns false when the csv first gets created and `olm.operatorNamespace` isn't set. 

Also, our update predicate only passes when there is an update to the csv AND an update to the meterdefinition housed in `marketplace.redhat.com/meterDefinition`. So if a csv is missing `olm.operatorNamespace` when our create predicate picks it up, it won't pass the create or update predicate and never get reconciled, unless you update the meterdef on the csv annotations

...this is why restarting the operator pod would pass the create predicate and the csv would get reconciled - when you restart the pod `olm.operatorNamespace` has had a few seconds to get set and is present on the csv annotations when the the `ClusterServiceVersionReconciler` runs.

### DIFF
--- a/v2/controllers/marketplace/clusterserviceversion_controller.go
+++ b/v2/controllers/marketplace/clusterserviceversion_controller.go
@@ -306,11 +306,16 @@ func (r *ClusterServiceVersionReconciler) reconcileMeterDefAnnotation(CSV *olmv1
 		return reconcile.Result{}, false, nil
 	}
 
-	if ns, ok := CSV.GetAnnotations()[olmNamespace]; ok && ns != CSV.GetNamespace() {
+	ns, ok := CSV.GetAnnotations()[olmNamespace]
+	if ok && ns != CSV.GetNamespace() {
 		reqLogger.Info("MeterDef is global and this CSV is not the head")
 		return reconcile.Result{}, false, nil
 	}
 
+	if !ok {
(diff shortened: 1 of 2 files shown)


dacleyra (approved):
lgtm

PR #314, opened 2022-03-18
Bugfix/workload resource results

dacleyra:
Problem

MeterDefinition with Updated spec.metrics
- Reflects old information on /metrics via prometheus_processor / prometheus_data
- Reflects old information on Status via status_processor

---

Resolved logic

dictionary gets Update event for MeterDefinition, standard delta.Update / cache.Update
  - Filter out non-Generational updates
  - StatusProcessor Status Update causes a lot of feedback looping
  - Unecessary processsing when Reporter job updates status
 
mdef_processor gets Update event for MeterDefinition, previously did nothing

Add a new onUpdate
  - Get the objects (pods/svc/pvc) indexed to the MeterDefinition
  - Delete the objects from the delta & indexstore (DeleteFromIndex doesn't try to delete from objectsSeen, although that may be a bad call in Delete anyways...)
	  - This Delete causes prometheus_processor & status_processor these objects should be removed from promdata and mdef Status
		- This is done because meterdefinition.spec.metrics entries could have been removed, and previous matching results were otherwise left accumulated
	- Call a meterDefinitionStore.Resync
	  - This will find object matches for the updated meterdefinition.spec.metrics
		- lookupCache was breaking the case when new meterdefinition.spec.metrics are added, don't use the cache
		- The Resync will go through the objectsSeen, and do an Add or Update for objects with FindObjectMatches results, in turn triggering prometheus_processor & status_processor to update /metrics & mdef status respectively
		
		
		
---

- remove bigcache results cache, a contributor to memory growth
- create a FindOwnerHelper that uses metadata client with informer & cache (v2/pkg/informedmetaclient/)
- the client starts an informer for resources that are looked up, such that the repeated calls hit the cache instead of rate limiting against the api-server
- clear Meterdefinition Status.WorkloadResources on (initial) Sync event, such that Status.WorkloadResources tracks correct resources if metric-state was down and missed delete events that should have removed resources from the list
- skip non-Generational Updates


zach-source (approved):
Changes look good to me.

zach-source (approved):
Tested the scale up/down scripts with this change and memory remained. Resources were removed from the endpoint. We still have some work to do there but it's much better.

PR #321, opened 2022-04-11
bugfix/thanosPrometheusMemory

zach-source:
- bugfix: cache expensive api creation
- bugfix: gogo protobuf security alert

Should function the same. Uses the operator local key instead of generating one on the API and having to refresh.


dacleyra (approved):
lgtm

PR #322, opened 2022-04-12
reduce cluster permissions

dacleyra:
- Remove cluster-scope get/list/watch for Secret by using cache.BuilderWithOptions
- Remove * permissions by replacing with cluster view role
  - view role does not get secrets
  - view role is dynamically updated to allow get/list/watch for 3rd party CRDs, negating the need for * permission


v2/config/rbac/role.yaml
```diff
@@ -65,11 +64,17 @@ rules:
...
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - secrets
+  verbs:
+  - '*'
```
zach-source, line 77: Should probably be specific with the verbs as well. If we can name the secrets as well the better.

zach-source (changes requested):
Minor nit on the secrets/configmap permission. Lets lock those down to specific get/watch/list and names if possible.

v2/config/rbac/role.yaml
dacleyra: fixed, forgot to add the file after the make generate

I will look into scoping to name as well, which will depend on if I can get fields.AndSelectors to work for the cache selector, which will be part of 24286

For now, the goal was to accomplish what was set out in the pr description

zach-source (dismissed):

zach-source (approved):

PR #325, opened 2022-04-15
fix: set marketplaceconfig accountid for iek

dacleyra:
- ensure that accountID is set on MarketplaceConfig when ibm-entitlement-key is used
- accountID is needed in the MarketplaceConfig spec, as consumed by marketplaceconfig_controller/marketplace_client to check registration status & unregister


mxpaspa (approved):
LGTM

zach-source (approved):

MJSiddu (approved):

PR #333, opened 2022-05-11
updated go to version 1.17.10

ab-ibm:
Update go version to 1.17.10 to address vulnerabilities

### DIFF
--- a/.github/workflows/branch_build.yml
+++ b/.github/workflows/branch_build.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.17
+        go-version: 1.17.10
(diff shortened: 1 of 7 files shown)


zach-source:
@dacleyra looks good to me, any concern about putting this in the release? We'll have to repackage it @MJSiddu heads up as well.

dacleyra:
I think we are still building airgap on 1.16, and that should be 1.17

https://github.com/redhat-marketplace/redhat-marketplace-operator/blob/develop/airgap/v2/Dockerfile#L3

zach-source:
@dacleyra Good catch Dan. Airgap could be a weird one cause of the dependencies it uses. May want to test thoroughly on 1.17 with it.

dacleyra:
updated by another pr, closing

PR #334, opened 2022-05-18
fix: csv filter

dacleyra:
This is for 24809 opened against RC 2.7.1

- Do not reconcile CSV deletes
  - We do not set finalizers on CSVs
  - Remove MeterDefinition delete logic, they are removed via OwnerRef
- Fix the csvFilter predicate
  - olmCopiedFromTag  / olm.copiedFrom is a Label not an Annotation, so the predicate was not filtering as intended
  - We were still reconciling for CSV copies, though reconcileMeterDefAnnotation() would catch these as copies and return anyways
 
 
There may still be an `Error Not Found` due to timing, as OLM fires a Status update against the primary CSV as an operator is being deleted. We can't ignore Status update events, as we consume Status in our reconcile.

### DIFF
--- a/v2/controllers/marketplace/clusterserviceversion_controller.go
+++ b/v2/controllers/marketplace/clusterserviceversion_controller.go
@@ -431,9 +431,15 @@ func csvFilter(metaNew metav1.Object) int {
 	return 0
 }
 
+func checkForUpdateToMdef(evt event.UpdateEvent) bool {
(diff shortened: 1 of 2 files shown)


MJSiddu (commented):
Looks good @dacleyra. Just 1 change as discussed on webex --> adding back the `checkForUpdateToMdef(evt)` check in update predicate

MJSiddu (approved):

PR #343, opened 2022-06-08
bugfix: buildah/podman support, dataservice 1.17

dacleyra:
- move dataservice to 1.17
- use explicit `--mount=` args, buildah/podman does not accept the shortcut
- fix the docker-push for base makefile
- fix the space var for utils.Makefile, it would otherwise leave a trailing comma in the arg which buildah/podman would fail on
- do not use a --push or --load with podman, the args are not available

_works for me_ with 
buildah 1.26.1
podman 4.1.0
rhel 9.0

### DIFF
--- a/airgap/v2/Dockerfile
+++ b/airgap/v2/Dockerfile
@@ -1,6 +1,6 @@
 ARG REGISTRY=quay.io/rh-marketplace
 
-FROM ${REGISTRY}/data-service-base:1.16-2 as dqlite-lib-builder
+FROM ${REGISTRY}/data-service-base:1.17 as dqlite-lib-builder
 ARG TARGETPLATFORM
 ARG TARGETARCH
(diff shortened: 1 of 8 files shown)


zach-source (approved):

zach-source:
Anyone else use podman want to test the change? I do not.

dacleyra:
lol, more or less as long as I have not broken the automated build, or anyone in their local environment

zach-source:
@dacleyra exactly - CI still works so I'm fine with it. Appears to only affect someone trying to build with podman. One thing to note is that podman probably can't built the multiarch images. You can run buildx on podman and that'll support it but that's whatever.

dacleyra:
merge after 2.7.1

mxpaspa (approved):

PR #344, opened 2022-06-08
bugfix: fix reports; fix operatorgroup filter

zach-source:
What's changed:

* Added missing role to marketplace-reporter service account
* Adjusted filtering to properly filter the operator group
  * Lots of tests added
* Namespace tracking enhancments
  * Tracking namespaces expanded to handle the reflector type (pod, service, etc)
  * If an all namespaces filter is detected for a type, it supersedes all other reflectors i.e. 1 global pod reflector vs 1 global pod reflect & namespace reflectors
  * Tons of testing to test different scenarios

### DIFF
--- a/metering/v2/pkg/engine/engine_test.go
+++ b/metering/v2/pkg/engine/engine_test.go
@@ -52,6 +52,7 @@ var _ = Describe("EngineTest", func() {
 			ctx,
 			pkgtypes.Namespaces{""},
 			scheme,
+			cfg,
 			managers.ClientOptions{
 				Namespace:    "",
 				DryRunClient: false,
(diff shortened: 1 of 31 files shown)


dacleyra:
- restored nonResourceURL for role to satisfy rhm-prom requirement
- restore * get/list/watch to metric-state to prevent forbidden errors
  - anticipate this fixes LS use case
  - revisit removal of * get/list/watch, and handling forbidden resources in the future

dacleyra:
- restore the nodes/metrics rbac to meterbase
  - the permission is needed for servicemonitor, a target was failing without, in the case of using rhm-prom, which I suspect is responsible for missing LS meterdefinition results
  - the servicemonitor uses bearerTokenSecret of the operator, using redhat-marketplace-service-account-token (this is a copy of the operator token, but is labeled, so it can be selected)
  - better would be to use the prometheus account token, which would better scope permission of the servicemonitor actor, but we would also need to solve watching that account/token and updating the servicemonitor, difficult, because there are no labels to use for selectors
  - best is to use bearerTokenFile, where the servicemonitor actor uses its own token with permissions, but I seem to recall this was not available in v4.6, which we still must support

dacleyra:
metering/v2/pkg/processors/prometheus_processor.go

Add some V(2) debugs

---

metering/v2/pkg/filter/lookup.go

Fix GetNamespaces to properly accumulate types for the namespaces map
Otherwise, types were not being added to the namespace watcher via the mdef_processor

---

metering/v2/pkg/processors/mdef_processor.go

Refactor the events to use the new store SyncByIndex functions

---

metering/v2/pkg/stores/store.go

Add was dropping adds to the indexStore that did not match a MeterDefinition
If a MeterDefinition was later added to a namespace, the objects would be missing in the cache to be matched
indexStore should always reflect the objects on the cluster, with the exception of status updates (same generation)

Fix incorrect locking in Update func

Move the Add object processing logic to addObj

Add Resync functions for updating objects when a meterdefinition change occurs
Resync() for entire cache
ResyncByIndex() to resync for objects that match a index: meterdefinition or from a namespace

Emit delta deletes when Resync or Update finds an object that no longer matches a meterdefinition

Deltas are processed by prometheus processor to update /metrics endpoint
Deltas are processed by status processor to update MeterDefinition status

PR #356, opened 2022-08-01
fix: build should use latest stable go

dacleyra:
For 
https://github.ibm.com/redhat-marketplace/psirt-automation/issues/1393
https://github.ibm.com/redhat-marketplace/psirt-automation/issues/1394

This supersedes #333 

- The workflows automatically retrieve the latest stable minor version with a given major version, the updates in that PR are unnecessary
- The base.Dockerfile is what actually gets used in the image, which was ultimately just 1.17.0 as set in the utils.Makefile
- This update will automatically pull the latest minor version if only major version is given, but GO_VERSION should also be set to a specific minor

### DIFF
--- a/airgap/v2/Dockerfile
+++ b/airgap/v2/Dockerfile
@@ -1,6 +1,6 @@
 ARG REGISTRY=quay.io/rh-marketplace
 
-FROM ${REGISTRY}/data-service-base:1.16-2 as dqlite-lib-builder
+FROM ${REGISTRY}/data-service-base:1.17 as dqlite-lib-builder
 ARG TARGETPLATFORM
 ARG TARGETARCH
 ARG TARGETOS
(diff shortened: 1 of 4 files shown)


ab-ibm:
@dacleyra I openeded similar PR some time ago https://github.com/redhat-marketplace/redhat-marketplace-operator/pull/333 I think also workflow should be updated to version 1.17

ab-ibm:
@dacleyra I think we have to upgrade to latest go version 1.17.13. I ran twistlock scan on updated images 2.7.1-beta.1121 and there are still some critical issues. Attached report
[twistlock-scan-results-20220802-085313-448400000-UTC-4E4CFA04.results.csv](https://github.com/redhat-marketplace/redhat-marketplace-operator/files/9240643/twistlock-scan-results-20220802-085313-448400000-UTC-4E4CFA04.results.csv)

PR #358, opened 2022-08-02
fix: ownerref and npe

dacleyra:
- handle err in marketplace client, if request fails, it leads to an NPE
- add a remove finalizer check for meterbase and razeedeployment
  - we no longer use the finalizer for cleanup of these objects, and use the ownerRef to have kube GC them
  - in an upgrade situation, if the finalizer was set by an older version, it would cause cleanup to fail with stuck finalizers (ex; upgrade 2.7.0 -> 2.7.1 -> uninstall)


mxpaspa (approved):
I went through the upgrade process, deleted marketplace config, and all resources were cleaned up

PR #369, opened 2022-09-23
fix: use self-built grpc-health-probe

dacleyra:
with untimely releases of grpc-health-probe release binary for CVEs, instead build/install using our (latest) version of golang


ab-ibm (approved):

ab-ibm:
Tested and scanned images by twistlock scan

PR #372, opened 2022-10-12
feat: support ocp v4.11

dacleyra:
- OCP version support v4.8-v4.11
- update to controller-runtime v0.12.3
- update to kube apis v0.24.0
- move from batchv1beta1 to batchv1.CronJob

- TODO: reconcileutils seems to begin to break at runtime, (meterbase) controllers need to move to using controllerutil instead


dacleyra:
intended changes are now in branch feature/splitOperator

PR #419, opened 2023-03-16
deployment controller

dacleyra:
create controller for each operator that specifically reconciles the loose bundle manifests 

problem: typically OLM will create the 4 bundle loose resources: secret, servicemonitor, service, configmap which enable protected /metrics endpoint. The Deployment specifically relies on the configmap volume to start. The secret can be prematurely destroyed if OLM creates the ServiceAccount before the Secret, as the Secret reconciler will delete an annotated secret if the serviceaccount does not exist (yet). The loose resources are not reconciled by OLM if they are modified/deleted.

### DIFF
--- a/deployer/v2/config/rbac/role.yaml
+++ b/deployer/v2/config/rbac/role.yaml
@@ -113,6 +113,19 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
+  - secrets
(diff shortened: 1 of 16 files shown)


ab-ibm (approved):
Verified on 4.9 cluster with default namespace redhat-marketplace

PR #441, opened 2023-05-16
operator-sdk 1.28.1, default api SA

dacleyra:


ab-ibm (approved):
code looks good but I'm not able to test on the cluster as images are not available probably due to branch name

PR #472, opened 2023-07-27
go-toolset 

mxpaspa:
For: https://github.ibm.com/symposium/track-and-plan/issues/30057


jiangchen2019 (dismissed):

dacleyra:
Since entrypoint is shared, this needs the changes for dataservice as well, and updating dataservice can wait on the grpc probe changes first

dacleyra:
Max Commits
- update Dockerfiles (not dataservice) to build on go-toolset to produce fips compliant binaries that use openssl, as per Red Hat fips cerfication guideline
  - remove base builder (not dataservice)
- update Dockerfiles to run fips-detect in startup process

Dan Commits
- Unify Dockerfiles (except dataservice), we can think about using/maintaining one or two files instead in the future
- base/dataservice.Dockerfile builds an enhanced go-toolset builder for building raft, dqlite, grpc probe and data-service with fips
- base/data-service-ubi-minimal.Dockerfile builds an enhanced ubi-minimal image which includes the built raft, dqlite and required libraries for data-service

jiangchen2019 (approved):

dacleyra (approved):

PR #512, opened 2023-09-28
Remove Rectest

mxpaspa:
For: https://github.ibm.com/symposium/track-and-plan/issues/30323


deployer/v2/controllers/marketplace/razeedeployment_controller_test.go
mxpaspa, line 1: @dacleyra the only test I couldn't get passing is the full uninstall. I left some objects commented out in case you need them for some reason

PR #676, opened 2024-09-13
feat: authorizeAccountCreation

dacleyra:
- Add spec.AuthorizeAccountCreation  to MarketplaceConfig
- Pass authorizeAccountCreation parameter to Reporter MarketplaceUploader
- use accounts endpoint to check existence


dacleyra:
this is ready to test against sandbox and prod

PR #726, opened 2025-03-17
[WIP, don't merge] initial drop of the segment-sdk compatibility changes

pgodowski:
API endpoint /v1/batch
bearer token in writeKey json payload

cc @dacleyra

### DIFF
--- a/datareporter/v2/Dockerfile
+++ b/datareporter/v2/Dockerfile
@@ -17,8 +17,8 @@ WORKDIR /src
 RUN --mount=type=bind,source=.,target=/src,z \
     --mount=type=cache,id=go-build,uid=1001,gid=0,target=/root/.cache/go-build \
     --mount=type=cache,id=mod,uid=1001,gid=0,target=/opt/app-root/src/go/pkg/mod \
-    cd ${path} && pwd && ls -l && echo ${exec} && echo ${bin} && \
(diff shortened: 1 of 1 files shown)


dacleyra:
Probably make the not valid logs .V(4)

https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/#authentication
Questions

How do you want to check writeKey? Checking for only presense of is not good.
Need to check for matching incoming writeKey
(Optional Mount) Secret-to-Env, Secret-to-File?
Do you want to remove the writeKey after auth?

Do you want to support Basic Auth too?

pgodowski:
It's work in progress, appreciate the comments.

> Probably make the not valid logs .V(4)

correct, working on this

> 
> https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/#authentication Questions
> 
> How do you want to check writeKey? Checking for only presense of is not good. Need to check for matching incoming 
writeKey (Optional Mount) Secret-to-Env, Secret-to-File? Do you want to remove the writeKey after auth?

I want to check whether writeKey is valid value of the bearer token - simply same token could be passed as Authentication: Bearer <token> or inside `writeKey` in the json payload, OR valid `Authentication: Basic base64(token)` option.
The code commited right now is useless  - will be rewritten.

> Do you want to support Basic Auth too?

Yes, see above.

